PR #147, opened 2021-11-11
1

serhii-dvorets:
https://serhii-dvorets.github.io/js_promises_practice_DOM/


danylo-onishchuk (changes requested):
The thirdPromise should be resolved with a message Third promise was resolved only after both left and right clicks happened

danylo-onishchuk (approved):

PR #182, opened 2022-03-07
js_promises_practice_DOM

ievgen-volkov:


src/scripts/main.js
```diff
@@ -1 +1,72 @@
...
+    resolve();
+  });
+});
+
+const rightClickPromise = new Promise(resolve => {
+  body.addEventListener('contextmenu', (e) => {
+    e.preventDefault();
+    resolve();
+  });
+});
```
vitaliy-martynyuk, line 41: these promises are not required, as you can use your `firstPromise` and `secondPromise` variables when handling `thirdPromise`

vitaliy-martynyuk (changes requested):

vitaliy-martynyuk (approved):

PR #189, opened 2022-04-14
Result

xpromt:
https://xpromt.github.io/js_promises_practice_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,67 @@
 'use strict';
+
+const body = document.querySelector('body');
+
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1 +1,67 @@
 'use strict';
+
+const body = document.querySelector('body');
+
+const message = (type, text) => {
```
vkryvytskyy, line 5:
```suggestion
const createMessage = (type, text) => {
```

vkryvytskyy (approved):
well done!

PR #197, opened 2022-05-22
solution

ShapovalDenys:
http://shapovaldenys.github.io/js_promises_practice_DOM/


src/scripts/main.js
```diff
@@ -1 +1,59 @@
...
+    }
+  });
+
+  body.addEventListener('contextmenu', () => {
+    onContextmenuClick = true;
+
+    if (onClick) {
+      resolve(message);
+    }
+  });
```
vkryvytskyy, line 42: there should be only one eventListener(mouseDown for example)

src/scripts/main.js
```diff
@@ -1 +1,59 @@
...
+const thirdPromise = new Promise((resolve) => {
+  const message = 'Third promise was resolved';
+  let onClick = false;
+  let onContextmenuClick = false;
+
+  body.addEventListener('click', () => {
+    onClick = true;
+
+    if (onContextmenuClick) {
+      resolve(message);
```
vkryvytskyy, line 32: resolve should happen only if both left and right buttons were clicked

vkryvytskyy (changes requested):

vkryvytskyy (approved):

PR #205, opened 2022-06-17
Develop

Alina-Yermolenko:
- [DEMO LINK](https://Alina-Yermolenko.github.io/js_promises_practice_DOM/)


src/scripts/main.js
```diff
@@ -1 +1,104 @@
...
+}
+
+promise1()
+  .then(result => {
+    const div = document.querySelector('body');
+
+    div.innerHTML
+    += `<div class="success"
+     data-qa="notification">
+     ${result}</div>`;
```
aliceblunt8, line 24: create a helper function, which creates a message and append it to body

src/scripts/main.js
```diff
@@ -1 +1,104 @@
 'use strict';
+
+function promise1() {
+  const resolver = (confirm1, reject1) => {
```
aliceblunt8, line 4: name these variables properly everywhere, without numbers

```suggestion
  const resolver = (resolve, reject) => {
```

src/scripts/main.js
```diff
@@ -1 +1,104 @@
...
+    document.addEventListener('contextmenu', () => {
+      right++;
+
+      if (left > 0 || right > 0) {
+        confirm2();
+      }
+    });
+  };
+
+  return new Promise(resolved);
```
aliceblunt8, line 57: >  The secondPromise should be:resolved only on left or right click in the document with a message Second promise was resolved

this should work, no need to check both
```
 document.addEventListener(('contextmenu'), (e) => {
    e.preventDefault();
    resolve(text);
  });

  document.addEventListener(('click'), (e) => {
    e.preventDefault();
    resolve(text);
  });
```

src/scripts/main.js
aliceblunt8: so that it looks something like this

```
firstPromise
  .then((message) => {
    createNotification(message, 'success');
  })
  .catch((message) => {
    createNotification(message, 'error');
  });
```

src/scripts/main.js
```diff
@@ -1 +1,104 @@
 'use strict';
+
+function promise1() {
```
aliceblunt8, line 3: let's make all promises variables and write all code inside Promise constructor
```suggestion
const firstPromise = new Promise((resolve, reject) => {
```

src/scripts/main.js
```diff
@@ -1 +1,104 @@
...
+
+    div.innerHTML += `<div 
+    class="success" 
+    data-qa="notification">
+    Second promise was resolved</div>`;
+  });
+
+function promise3() {
+  let left = 0;
+  let right = 0;
```
aliceblunt8, line 72: give these variables a boolean value (true) when they are clicked

```suggestion
  let left;
  let right;
```

aliceblunt8 (changes requested):

PR #263, opened 2022-09-15
all done but in readme needs only left click to success first promise…

Yevhen-Pavlovskyi:
in readme needs only left click to success first promise, but test requires 0 || 1 || 2 butto success first promise


SanyaBratashchuk (approved):
Interesting solution

PR #281, opened 2022-10-04
solution

vitalii-babiienko:
[DEMO LINK](https://vitalii-babiienko.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,68 @@
 'use strict';
+
+const body = document.body;
+
+const firstPromise = new Promise((resolve, reject) => {
+  body.addEventListener('click', () => {
+    resolve('First promise was resolved');
+  });
+
+  setTimeout(() => {
+    reject(new Error('First promise was rejected'));
+  }, 3000);
+});
+
+const secondPromise = new Promise((resolve, reject) => {
+  body.addEventListener('mousedown', e => {
+    if (e.button === 0 || e.button === 2) {
+      resolve('Second promise was resolved');
+    }
+  });
+});
+
+const thirdPromise = new Promise((resolve, reject) => {
+  let leftClick = false;
+  let rightClick = false;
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1 +1,68 @@
...
+      rightClick = true;
+    }
+
+    if (leftClick && rightClick) {
+      resolve('Third promise was resolved');
+    }
+  });
+});
+
+function success(message) {
```
qoa11a, line 42: success and error functions are alike. I recommend to create one function showMessage(message, class) {}

src/scripts/main.js
```diff
@@ -1 +1,68 @@
...
+    }
+  });
+});
+
+const thirdPromise = new Promise((resolve, reject) => {
+  let leftClick = false;
+  let rightClick = false;
+
+  body.addEventListener('mousedown', e => {
+    if (e.button === 0) {
```
qoa11a, line 28: Your solution is correct but I believe there's better way. Read about this event
https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event

qoa11a (changes requested):

src/scripts/main.js
```diff
@@ -1 +1,53 @@
...
+    resolve('Second promise was resolved');
+  });
+});
+
+const thirdPromise = new Promise((resolve, reject) => {
+  body.addEventListener('click', () => {
+    body.addEventListener('contextmenu', () => {
+      resolve('Third promise was resolved');
+    });
+  });
```
SanyaBratashchuk, line 32: А якщо спершу натиснути праву кнопку миші?  Клік не зарахується

SanyaBratashchuk (changes requested):

marina-tilniak (approved):

PR #285, opened 2022-10-07
Solution by DK

DimaK123:
[DEMO LINK](https://DimaK123.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,60 @@
 'use strict';
+
+const body = document.querySelector('body');
+const logo = document.querySelector('.logo');
+
+const promise1 = new Promise((resolve, reject) => {
+  logo.addEventListener('click', () => {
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1 +1,60 @@
 'use strict';
+
+const body = document.querySelector('body');
+const logo = document.querySelector('.logo');
+
+const promise1 = new Promise((resolve, reject) => {
+  logo.addEventListener('click', () => {
```
qoa11a, line 7: Add event listener on document, not on logo.

qoa11a (changes requested):
Fix it everywhere.

src/scripts/main.js
DimaK123: Done

polosanya (approved):
Good job!

PR #297, opened 2022-10-31
commit message

11Olya112:


src/scripts/main.js
```diff
@@ -1 +1,47 @@
...
+    resolve('First promise was resolved');
+  });
+
+  setTimeout(() => {
+    reject(new Error('First promise was rejected'));
+  }, 3000);
+});
+
+const promise2 = new Promise((resolve) => {
+  document.addEventListener('mousedown', () => {
```
Bohdan-Maksymiuk, line 14: now second promise resolved not only on left click

src/scripts/main.js
```diff
@@ -1 +1,47 @@
...
+    resolve('Second promise was resolved');
+  });
+});
+
+const promise3 = new Promise((resolve) => {
+  document.addEventListener('click' && 'contextmenu', () => {
```
Bohdan-Maksymiuk, line 20: how does it work?
you pass 'click' && 'contextmenu' returns  'contextmenu', so you set listener for `contextmenu` only
it doesn't suit the task description

Bohdan-Maksymiuk (changes requested):
https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons
https://javascript.info/mouse-events-basics

polosanya (changes requested):
Third promise doesn't resolve on left and right click if first promise was rejected
image
image

loralevitska (changes requested):
Third promise doesn't resolve on left and right click in both case 
image

image

11Olya112:
я оновила demo link

src/scripts/main.js
```diff
@@ -1 +1,60 @@
...
+    appendDiv(res, 'success success__1');
+  })
+  .catch((res) => {
+    appendDiv(res, 'warning');
+  });
+
+promise2
+  .then((res) => {
+    appendDiv(res, 'success success__2');
+  });
```
Anastasiia-Svintsova, line 55:
```suggestion
promise2.then((res) => {
  appendDiv(res, 'success success__2');
});
```

src/scripts/main.js
```diff
@@ -1 +1,60 @@
...
+
+promise3
+  .then((res) => {
+    appendDiv(res, 'success success__3');
+  });
```
Anastasiia-Svintsova, line 60:
```suggestion
promise3.then((res) => {
  appendDiv(res, 'success success__3');
});
```

src/styles/main.scss
```diff
@@ -22,3 +22,39 @@ body {
...
+
+  &__2 {
+    right: 50px;
+  }
+
+  &__3 {
+    bottom: 50px;
+    right: 50px;
+  }
+}
```
Anastasiia-Svintsova, line 49: Look how many times you use 50px. It would be better to create a variable for it

Anastasiia-Svintsova (approved):

PR #299, opened 2022-11-02
Solution

anastasia-babko:
[DEMO LINK](https://anastasia-babko.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,66 @@
 'use strict';
+
+document.addEventListener('contextmenu', (e) => {
+  e.preventDefault();
+});
+
+const firstPromise = new Promise((resolve, reject) => {
+  document.addEventListener('mousedown', (e) => {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1 +1,66 @@
...
+  document.addEventListener('mousedown', (e) => {
+    if (e.button >= 0) {
+      resolve('First promise was resolved');
+    }
+  });
+
+  setTimeout(() => reject(new Error('First promise was rejected')), 3000);
+});
+
+const secondPromise = new Promise((resolve, reject) => {
```
DarkMistyRoom, line 17:
```suggestion
const secondPromise = new Promise(resolve => {
```

DarkMistyRoom (changes requested):
Your first promise resolves on both clicks, but should - only on the left one

SanyaBratashchuk (approved):

PR #305, opened 2022-11-07
Solution

vitaliidox:
https://vitaliidox.github.io/js_promises_practice_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,59 @@
 'use strict';
+
+const promise1 = new Promise((resolve, reject) => {
+  document.addEventListener('click', () => {
+    resolve('First promise was resolved');
+  });
+
+  setTimeout(() => {
+    reject(new Error('First promise was rejected'));
+  }, 3000);
+})
+  .then(data => {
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1 +1,59 @@
...
+    </div>
+    `);
+  })
+  .catch(err => {
+    document.body.innerHTML += `
+    <div data-qa="notification" class="warning">
+      ${err}
+    </div>
+    `;
+  });
```
vlad-demchuk, line 25: call it separately
image

src/scripts/main.js
```diff
@@ -1 +1,59 @@
...
+    <div data-qa="notification" class="success">
+      ${data}
+    </div>
+    `);
+  })
+  .catch(err => {
+    document.body.innerHTML += `
+    <div data-qa="notification" class="warning">
+      ${err}
+    </div>
```
vlad-demchuk, line 23: better to use `textContent`, `append, prepend...` or ` insertAdjacentHTML`

vlad-demchuk (changes requested):

vlad-demchuk (approved):

PR #310, opened 2022-11-11
three promises was created

iPanda177:
[DEMO LINK](https://iPanda177.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,65 @@
 'use strict';
+
+const root = document.querySelector('body');
+
+const firstPromise = new Promise((resolve, reject) => {
+  root.addEventListener('click', () => (
+    resolve('First promise was resolved')
+  ));
+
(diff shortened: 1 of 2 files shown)


Anna-Tkachenko (changes requested):
Check the process of resolving the third promise, it doesn't work for me👀

iPanda177:
strange, works fine for me... maybe that's because of difference of mac mouse? add listener for another extra button

src/scripts/main.js
```diff
@@ -1 +1,65 @@
...
+  root.addEventListener('contextmenu', () => (
+    resolve('Second promise was resolved')
+  ));
+});
+
+const thirdPromise = new Promise((resolve) => {
+  let left = false;
+  let right = false;
+
+  root.addEventListener('mouseup', (ev) => {
```
Anna-Tkachenko, line 27: you can also handle events, that will occur on left and right clicks, instead of handling separate buttons

PR #314, opened 2022-11-12
Added a task solution

oleksusov:
[DEMO LINK](https://oleksusov.github.io/js_promises_practice_DOM/)


polosanya (changes requested):
The logo should not disappear. Also, first and second promises don't work correctly. 
Please check yourself with the task description and try to fix it
image

polosanya (changes requested):
Third promise still doesn't work correctly. 
promises
Add some variable for left and right click and resolve third promise if both of them are true.
Also, try not to move logo at all

polosanya (approved):
Well done!

PR #321, opened 2022-11-26
Implemented the solution

zynkd:


src/scripts/main.js
```diff
@@ -1 +1,72 @@
...
+      resolve('Third promise was resolved');
+    }
+  });
+});
+
+thirdPromise.then((message) => {
+  body.insertAdjacentHTML(
+    'beforeend',
+    `<div data-q="notification" class="msg msg--success3">${message}</div`
+  );
```
polosanya, line 71: You have written this three times. Don't repeat yourself, create a function for adding messages

polosanya (changes requested):
Almost perfect! Please check my comment

src/scripts/main.js
zynkd: Hi, should be better now. Let me know. Thanks!

src/scripts/main.js
```diff
@@ -1 +1,58 @@
...
+}
+
+function getErrorMessage(message) {
+  return body.insertAdjacentHTML(
+    'beforeend',
+    `<div data-q="notification" class="
+     message message--warning">${message.message}</div>`
+  );
+}
```
qoa11a, line 18: These two function are alike. Try to move it to a single function, than inside this function you can decide if to add 'message-warning' class.

Function's usage might look like this.
`getMessage('Some message', 'success');`
Hint: you may use 2nd argument that will be type of message.

src/scripts/main.js
```diff
@@ -1 +1,58 @@
...
+  document.addEventListener('click', () => {
+    leftClick = true;
+
+    if (leftClick && rightClick) {
+      resolve('Third promise was resolved');
+    }
+  });
+
+  document.addEventListener('contextmenu', (e) => {
+    e.preventDefault(); // prevent right-click popup menu
```
qoa11a, line 47: Remove unnecessary comments. It's obvious that e.preventDefault() on contextmenu prevents opening of context menu.

qoa11a (changes requested):
Good job! 👍 
Just few comments.

src/scripts/main.js
zynkd: Hi, I've changed the code now - let me know. Thanks!

amitpatiljc (approved):
Nice 🔥

PR #322, opened 2022-11-26
solved

Deniso4ka23:
[DEMO LINK](https://Deniso4ka23.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,57 @@
 'use strict';
+
+const root = document.body;
+
+// firstPromise
+new Promise((resolve, reject) => {
+  document.addEventListener('click', () => {
+    resolve();
+  });
+
+  setTimeout(() => {
+    reject(new Error());
+  }, 3000);
+}).then(() => {
+  root.insertAdjacentHTML('beforeend', `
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1 +1,57 @@
 'use strict';
+
+const root = document.body;
+
+// firstPromise
+new Promise((resolve, reject) => {
```
polosanya, line 6:
```suggestion
const firstPromise = new Promise((resolve, reject) => {
```
Just create variables for promises and won't need the comments

src/scripts/main.js
```diff
@@ -1 +1,57 @@
...
+  document.addEventListener('click', () => {
+    resolve();
+  });
+
+  document.addEventListener('contextmenu', e => {
+    e.preventDefault();
+    resolve();
+  });
+}).then(() => {
+  root.insertAdjacentHTML('beforeend', `
```
polosanya, line 35: You do the same in all cases. Create function for creating messages (success\error)

polosanya (changes requested):
Good job! Please check my comments to improve your code a bit

PR #324, opened 2022-11-29
Promises practice has been solved

Oleksandr-Antonenko:
- [DEMO LINK](https://Oleksandr-Antonenko.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,71 @@
 'use strict';
+
+const notifications = document.createElement('div');
+
+notifications.classList.add('notifications');
+document.body.appendChild(notifications);
+
+const generateNotification = (className, message) => {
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1 +1,71 @@
...
+    }
+
+    if (leftClick || rightClick) {
+      resolve('Second promise was resolved');
+    }
+  });
+});
+
+const thirdPromise = new Promise((resolve) => {
+  let leftClick, rightClick;
```
vlad-demchuk, line 46: set value for these variables and create it separately for clarity.

vlad-demchuk (changes requested):

vlad-demchuk (approved):

PR #325, opened 2022-12-04
3 promises

miamarichka:
[DEMO LINK] https://miamarichka.github.io/js_promises_practice_DOM/


liliia-livandovskaya (approved):
Great job! But it would be better to add different styles of success and error

PR #327, opened 2022-12-07
Solution

UmizDemud:
- [DEMO LINK](https://UmizDemud.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,81 @@
 'use strict';
+
+let leftClicked = false;
+let rightClicked = false;
+
+const firstPromise = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(Error('First promise was rejected'));
+  }, 3000);
+
+  document.body.addEventListener('click', e => {
+    leftClicked = true;
+    document.body.removeEventListener('click', this);
+    resolve('First promise was resolved');
+  });
+});
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1 +1,81 @@
 'use strict';
+
+let leftClicked = false;
+let rightClicked = false;
```
mgrinko, line 4: Don't use global boolean flags
create a promise per each event 
and combine them with Promise.all

src/scripts/main.js
```diff
@@ -1 +1,81 @@
...
+
+const firstPromise = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(Error('First promise was rejected'));
+  }, 3000);
+
+  document.body.addEventListener('click', e => {
+    leftClicked = true;
+    document.body.removeEventListener('click', this);
```
mgrinko, line 13: you could add once listener instead of removing it manually
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#parameters

see options -> once

mgrinko (changes requested):

src/scripts/main.js
UmizDemud: thanks!

naziamoff (approved):
👍

PR #336, opened 2022-12-16
Solution

anton-sosonskyi:
[DEMO](https://anton-sosonskyi.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,57 @@
 'use strict';
+
+const createMessage = function(message, statusClass) {
+  const div = document.createElement('div');
+
+  div.dataset.qa = 'notification';
+  div.className = statusClass;
+  div.textContent = message;
+
+  document.body.append(div);
+};
+
+const firstPromise = new Promise((resolve, reject) => {
+  document.addEventListener('mousedown', (e) => {
+    if (e.button === 0) {
+      resolve('First promise was resolved');
+    }
+  });
+
+  setTimeout(() => {
+    reject(new Error('First promise was rejected'));
+  }, 3000);
+});
+
+const secondPromise = new Promise((resolve, reject) => {
+  document.addEventListener('mousedown', (e) => {
+    const mouseButton = e.button;
+
+    if (mouseButton === 0 || mouseButton === 2) {
+      resolve('Second promise was resolved');
+    }
+  });
+});
+
+const thirdPromise = new Promise((resolve, reject) => {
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1 +1,57 @@
...
+});
+
+const thirdPromise = new Promise((resolve, reject) => {
+  document.addEventListener('mousedown', (e) => {
+    if (e.button === 0) {
+      document.addEventListener('mousedown', (click) => {
+        if (click.button === 2) {
+          resolve('Third promise was resolved');
+        }
+      });
```
SanyaBratashchuk, line 42: It will not work if I start with right click

SanyaBratashchuk (changes requested):

src/scripts/main.js
```diff
@@ -1 +1,65 @@
...
+    }
+
+    if (leftClick && rightClick) {
+      resolve('Third promise was resolved');
+    }
+  });
+});
+
+firstPromise
+  .then(result => createMessage(result, `success`))
```
OlEzhA0, line 56: you create the same callback for each `.then` and for each `.catch` so you can add some functions and pass it here

src/scripts/main.js
```diff
@@ -1 +1,65 @@
...
+    }
+  });
+
+  setTimeout(() => {
+    reject(new Error('First promise was rejected'));
+  }, 3000);
+});
+
+const secondPromise = new Promise((resolve, reject) => {
+  document.addEventListener('mousedown', (e) => {
```
OlEzhA0, line 26: you can listen 'contextmenu' for right clicks and prevent default popup

OlEzhA0 (changes requested):
Hey! I left really small improvements so you can check it

src/scripts/main.js
```diff
@@ -1 +1,72 @@
...
+
+document.addEventListener('contextmenu', (e) => {
+  e.preventDefault();
+});
+
+const firstPromise = new Promise((resolve, reject) => {
+  document.addEventListener('mousedown', (e) => {
+    if (e.button === 0) {
+      resolve('First promise was resolved');
+    }
```
DarkMistyRoom, line 24: You can listen `click`, then you don't need to check the button

DarkMistyRoom (changes requested):
You can just fix the comment, but think about using separate promises for left and right clicks, then you can avoid all `if`s

vlad-demchuk (approved):

PR #343, opened 2022-12-26
Solution 32

mary-kucher:
[DEMO LINK](https://mary-kucher.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,60 @@
 'use strict';
+
+const body = document.querySelector('body');
+
+const firstPromise = new Promise((resolve, reject) => {
+  document.addEventListener('mousedown', () => {
+    resolve('First promise was resolved');
+  });
+
+  setTimeout(() => {
+    reject(new Error('First promise was rejected'));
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1 +1,64 @@
 'use strict';
+
+const firstPromise = new Promise((resolve, reject) => {
+  document.addEventListener('mousedown', () => {
+    resolve('First promise was resolved');
```
TarasHoliuk, line 5: it should be resolved only on the **left** button click:
image

TarasHoliuk (changes requested):

src/scripts/main.js
mary-kucher: Yes, I saw it. But tests sad me that not right ((

image


> it should be resolved only on the **left** button click: image

TarasHoliuk: I see
Then it's mismatching between tests and description - not your fault
So I'll approve the PR

TarasHoliuk (approved):

PR #347, opened 2023-01-01
vol_1

RudBeee:
https://RudBeee.github.io/js_promises_practice_DOM

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,94 @@
 'use strict';
+
+const doc = document.querySelector('html');
+const body = document.querySelector('body');
+let count = 0;
+
+function creatPromis() {
+  const resolver = (success, warning) => {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1 +1,94 @@
 'use strict';
+
+const doc = document.querySelector('html');
+const body = document.querySelector('body');
+let count = 0;
+
+function creatPromis() {
+  const resolver = (success, warning) => {
```
tetiana-ivanchenko, line 8:
```suggestion
function creatPromise() {
  const resolver = (resolve, reject) => {
```

tetiana-ivanchenko (changes requested):
image
your 3 promise resolve after two left clicks

nazarmatsevych (approved):

PR #349, opened 2023-01-03
JS promises practice DOM

partnersinbahamas:
[DEMO LINK](https://partnersinbahamas.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,48 @@
 'use strict';
+
+const body = document.querySelector('body');
+
+function notification(message, classes) {
+  body.insertAdjacentHTML('beforeend', `
+      <div class="${classes}"data-qa="notification">${message}</div>
+    `);
+};
+
+const firstPromise = new Promise((resolve, reject) => {
+  body.addEventListener('click', () => {
+    resolve();
+  });
+
+  setTimeout(() => {
+    reject(new Error());
+  }, 3000);
+});
+
+firstPromise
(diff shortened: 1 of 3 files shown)


partnersinbahamas:
Sorry, I don't understand how to fix my tests, cause a don't know how to click on the right button on Mac, I mean, I know that I must make a click with a two fingers on the touch board, but its only opens a menu((

src/scripts/main.js
```diff
@@ -1 +1,48 @@
...
+
+secondPromise
+  .then(() => notification('Second promise was resolved', 'success'));
+
+const thirdPromises = new Promise((resolve, reject) => {
+  body.addEventListener('click', (events) => {
+    const leftButton = events.button === 0;
+    const rightButton = events.button === 2;
+
+    if (leftButton && rightButton) {
```
witflash, line 41: This check will never be true because a single event has a single button. You have to handle every click and save it outside the eventListener. For example:
```
  let isLeftClick = false;
  let isRightClick = false;
  
  body.addEventListener('click', event => ...
```

witflash (changes requested):
Good job, but you have a problem with your third Promise.

partnersinbahamas:
Done, but still have a problem with rbutton, and at third promise some test haven't done.

witflash (changes requested):
Hey!
I see your trouble with the right button.
Look. You can use `mousedown` event instead the `click` event. And all will work fine for you.

nazarmatsevych (approved):

PR #355, opened 2023-01-07
Solution

OksanaShvets1:
- [DEMO LINK](https://OksanaShvets1.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,69 @@
 'use strict';
+
+const firstPromise = new Promise((resolve, reject) => {
+  document.querySelector('.logo').addEventListener('click', function() {
+    resolve();
+  });
+
+  setTimeout(() => reject(new Error('promise was rejected')), 3000);
+});
+
+firstPromise.then(() => {
+  const div = document.createElement('div');
+
+  div.setAttribute('data-qa', 'notification');
+  div.className = 'success';
+  div.innerText = 'First promise was resolved';
+  document.body.append(div);
+});
+
+firstPromise.catch(() => {
+  const div = document.createElement('div');
+
+  div.setAttribute('data-qa', 'notification');
+  div.className = 'warning';
+  div.innerText = 'First promise was rejected';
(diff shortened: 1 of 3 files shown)


tetiana-ivanchenko (changes requested):
image

src/scripts/main.js
```diff
@@ -1 +1,69 @@
...
+});
+
+firstPromise.catch(() => {
+  const div = document.createElement('div');
+
+  div.setAttribute('data-qa', 'notification');
+  div.className = 'warning';
+  div.innerText = 'First promise was rejected';
+  document.body.append(div);
+});
```
vlad-demchuk, line 27:
```suggestion
firstPromise
  .then(() => {
  const div = document.createElement('div');

  div.setAttribute('data-qa', 'notification');
  div.className = 'success';
  div.innerText = 'First promise was resolved';
  document.body.append(div);
})
  .catch(() => {
  const div = document.createElement('div');

  div.setAttribute('data-qa', 'notification');
  div.className = 'warning';
  div.innerText = 'First promise was rejected';
  document.body.append(div);
});
```

vlad-demchuk: you can do it in a chain

also create a separate function for a notification and pass needed data there

vlad-demchuk (changes requested):

src/scripts/main.js
```diff
@@ -1 +1,62 @@
...
+
+firstPromise
+  .then(() => {
+    printMessage('First promise was resolved', 'success');
+  })
+
+  .catch(() => {
+    printMessage('First promise was rejected', 'warning');
+  });
```
vlad-demchuk, line 18:
```suggestion
firstPromise
  .then(() => {
    printMessage('First promise was resolved', 'success');
  })
  .catch(() => {
    printMessage('First promise was rejected', 'warning');
  });
```

vlad-demchuk (approved):

PR #356, opened 2023-01-07
three promises

KateRiaz:
[demo](https://kateriaz.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,72 @@
 'use strict';
+
+const body = document.querySelector('body');
+const firstPromise = new Promise((resolve, reject) => {
+  document.addEventListener('click', () => {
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1 +1,72 @@
...
+  document.addEventListener('click', () => {
+    resolve('First promise was resolved');
+  });
+
+  setTimeout(() => {
+    reject(new Error('First promise was rejected'));
+  }, 3000);
+});
+
+function success(result) {
```
polosanya, line 14:
```suggestion
function handleSuccess(result) {
```
Let's get used to name function according to these rules:
image

polosanya (approved):
Good job!

PR #357, opened 2023-01-08
task resolved

oleksiy-uchaev:
- [DEMO LINK](https://oleksiy-uchaev.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,41 @@
 'use strict';
+
+const addText = (number, isError) => `
+  ${number} promise was ${isError ? 'rejected' : 'resolved'}
+`;
+
+const firstPromise = new Promise((resolve, reject) => {
+  document.addEventListener('mousedown', (e) => {
+    e.button === 0 && resolve(addText('First'));
+  });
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1 +1,41 @@
 'use strict';
+
+const addText = (number, isError) => `
```
loralevitska, line 3: In fact, this variable `number ` does not store a number, but the name of the promise, rename it.

src/scripts/main.js
```diff
@@ -1 +1,41 @@
...
+  ${number} promise was ${isError ? 'rejected' : 'resolved'}
+`;
+
+const firstPromise = new Promise((resolve, reject) => {
+  document.addEventListener('mousedown', (e) => {
+    e.button === 0 && resolve(addText('First'));
+  });
+  setTimeout(() => reject(addText('First', 1)), 3000);
```
loralevitska, line 11: In this function, you have a Boolean type as the second parameter `isError`, so pass that type there `true`/`false`

loralevitska (changes requested):
Almost done)

qoa11a (approved):

PR #358, opened 2023-01-08
Solution

IVxVI:
[DEMO LINK](https://ivxvi.github.io/js_promises_practice_DOM/)


src/scripts/main.js
```diff
@@ -1 +1,78 @@
...
+      resolve(successMessage3);
+    });
+  });
+
+  logo.addEventListener('contextmenu', (ev) => {
+    logo.addEventListener('click', () => {
+      ev.preventDefault();
+      resolve(successMessage3);
+    });
+  });
```
SanyaBratashchuk, line 51: you can create 2 variables leftClicked and rightClicked for example, with false value by default and change theirs values when click happens. And check if values of these variables are true if yes resolve promise

SanyaBratashchuk (changes requested):

src/scripts/main.js
IVxVI: I couldn`t handle it without creating two eventListeners for left-right clicks and two right-left clicks, so I decided to use 'mousedown' with buttons check, cuz it looks more optimized =)

SanyaBratashchuk (approved):

PR #367, opened 2023-01-16
solution

divineretribution1:
[DEMO LINK](https://divineretribution1.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,55 @@
 'use strict';
+
+const body = document.querySelector('body');
+
+body.addEventListener('contextmenu', (e) => {
+  e.preventDefault();
+});
+
+const promise1 = new Promise((resolve, reject) => {
+  body.addEventListener('mousedown', ev => {
+    if (ev.button === 0) {
+      resolve();
+    };
+  });
+
+  setTimeout(() => reject(new Error()), 3000);
+});
+
+const promise2 = new Promise((resolve) => {
+  body.addEventListener('mousedown', ev => {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1 +1,55 @@
...
+
+const promise3 = new Promise((resolve) => {
+  body.addEventListener('mousedown', ev1 => {
+    body.addEventListener('mousedown', ev2 => {
+      if ((ev1.button === 0 && ev2.button === 2)
+      || (ev1.button === 2 && ev2.button === 0)) {
+        resolve();
+      };
+    });
+  });
```
loralevitska, line 35: It's better avoid nested listeners. Create two boolean variables `rightClick` `leftClick` and set them true value if it was right or left click on body. Then just check this 2 variables and if theese variables are right resolve 3d promise

src/scripts/main.js
```diff
@@ -1 +1,55 @@
...
+});
+
+promise1.then(() => {
+  body.insertAdjacentHTML('beforeend',
+    '<div data-qa="notification">First promise was resolved</div>');
+})
+  .catch(() => {
+    body.insertAdjacentHTML('beforeend',
+      '<div data-qa="notification">First promise was rejected</div>');
```
loralevitska, line 44: combine repeatable code in variable or helper function

loralevitska (changes requested):
Almost done)

qoa11a (approved):
good job 👍

PR #368, opened 2023-01-16
Solution

Vlad-Fedorishchev:
Demo Link  -  https://Vlad-Fedorishchev.github.io/js_promises_practice_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,71 @@
 'use strict';
+
+const bodyElement = document.querySelector('body');
+
+bodyElement.addEventListener('contextmenu', (action) => {
+  action.preventDefault();
+});
+
+// Messages classes-------------------------------------------------------------
+const messageClass = 'message';
+const secondMessageClass = 'message second-message';
+const thirdMessageClass = 'message third-message';
+const errorMessageClass = 'message error-message';
+
+// Messages types---------------------------------------------------------------
+const resolvedMessage = 'First promise was resolved';
+const rejectedMessage = 'First promise was rejected';
+const secondResolveMessage = 'Second promise was resolved';
+const thirdPromiseMessage = 'Third promise was resolved';
+
+// Show messages function-------------------------------------------------------
+function showMessage(classType, MessageType) {
+  bodyElement.insertAdjacentHTML('beforeend', `
+        <div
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1 +1,71 @@
...
+const messageClass = 'message';
+const secondMessageClass = 'message second-message';
+const thirdMessageClass = 'message third-message';
+const errorMessageClass = 'message error-message';
+
+// Messages types---------------------------------------------------------------
+const resolvedMessage = 'First promise was resolved';
+const rejectedMessage = 'First promise was rejected';
+const secondResolveMessage = 'Second promise was resolved';
+const thirdPromiseMessage = 'Third promise was resolved';
```
loralevitska, line 19: You can combine it in helper function that returns string `${...} promise was ${...}`
And same do with messageClass variables

src/scripts/main.js
```diff
@@ -1 +1,71 @@
...
+// Realizing thirdPromise-------------------------------------------------------
+const thirdPromise = new Promise((resolve) => {
+  bodyElement.addEventListener('mousedown', (actionOne) => {
+    bodyElement.addEventListener('mousedown', (actionTwo) => {
+      if ((actionOne.button === 0 && actionTwo.button === 2)
+            || (actionOne.button === 2 && actionTwo.button === 0)) {
+        resolve();
+      }
+    });
+  });
```
loralevitska, line 68: Try to avoid nesting listeners. Check it with two variables in global scope

loralevitska (changes requested):
Cool 😎 Some comments to fix:

1. Remove all comments. Try it write obvious code that explain all we need without comments)

src/scripts/main.js
```diff
@@ -1 +1,66 @@
...
+  let leftClickCheck = false;
+  let rightClickCheck = false;
+
+  bodyElement.addEventListener('mousedown', (action) => {
+    switch (action.button) {
+      case 0:
+        leftClickCheck = true;
+        break;
+      case 2:
+        rightClickCheck = true;
```
polosanya, line 56: Don't forget to add a default case to avoid possible errors

polosanya (approved):

PR #375, opened 2023-01-21
solution

1stIndominus:
[DEMO LINK](https://1stIndominus.github.io/js_promises_practice_DOM/)


Anastasiia-Svintsova (changes requested):
- The task: The thirdPromise should be resolved with a message Third promise was resolved only after both left and right clicks happened
But for now, the third promise is resolved after two right clicks also

https://user-images.githubusercontent.com/99403198/213869919-4eebba47-a38c-4b54-bc72-e4c71d48771a.mov

polosanya (approved):

PR #382, opened 2023-01-27
solution

anna-akhalmosulishvili:
https://anna-akhalmosulishvili.github.io/js_promises_practice_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,83 @@
 'use strict';
+
+const body = document.querySelector('body');
+
+function createNotification(state, content) {
+  const notification = document.createElement('div');
+
+  body.appendChild(notification);
+  notification.setAttribute('data-qa', 'notification');
+  notification.textContent = content;
+
+  if (state === 'success') {
+    notification.classList.add('success');
+  } else {
+    notification.classList.add('warning');
+  }
+}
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1 +1,83 @@
...
+
+    body.addEventListener('contextmenu', e => {
+      e.target.addEventListener('click', () => {
+        success(`Third promise was resolved`);
+      });
+    });
+  };
+
+  return new Promise(resolver);
+}
```
loralevitska, line 63: Better use 1 listener on body and have two boolean variables (for left and right click). When two variables wiil be truthy resolve promise

src/scripts/main.js
```diff
@@ -1 +1,83 @@
...
+
+  if (state === 'success') {
+    notification.classList.add('success');
+  } else {
+    notification.classList.add('warning');
+  }
+}
+
+function createFirstPromise() {
+  const resolver = (success, error) => {
```
loralevitska, line 20: Fix all cases please
```suggestion
  const resolver = (resolve, reject) => {
```

loralevitska (changes requested):
Almost done

qoa11a (approved):

PR #393, opened 2023-02-07
solution

vladyslav-krainyk:
[DEMO LINK](https://vladyslav-krainyk.github.io/js_promises_practice_DOM/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Add `success` and `error` handlers to each promise to show `<div data-qa="notifi
 
 ## Instructions
 1. Replace `<your_account>` with your Github username in the link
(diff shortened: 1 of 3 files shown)


slavanosatskyi (changes requested):
very nice work :) the only thing that I would suggest to change is to properly handle right click button (check event parameter for this case). There are keyboards with contextmenu button which can trigger this event in browser as well and consider to use another event (mouse down for example)

SanyaBratashchuk (changes requested):
comment ⏫

Volodymyr-Melnyk-Fron10der (approved):
Great job!

PR #395, opened 2023-02-09
solution

anastasia-radchencko:
https://anastasia-radchencko.github.io/js_promises_practice_DOM/


Volodymyr-Melnyk-Fron10der (approved):
Great! But as you see the text 'Second promise was resolved' is appear first because event `mousedown` is triggered before `click`. So I would handle `mousedown` for first promise as well. Anyway all checkpoints are completed so PR is approved

PR #396, opened 2023-02-12
task was done

vanyachyzh:
- [DEMO LINK](https://vanyachyzh.github.io/js_promises_practice_DOM/)


src/scripts/main.js
```diff
@@ -1 +1,72 @@
 'use strict';
+
+const BODY = document.body;
```
OlEzhA0, line 3:
```suggestion
const body = document.body;
```

src/scripts/main.js
```diff
@@ -1 +1,72 @@
...
+
+function showSuccess(result) {
+  const MESSAGE = document.createElement('div');
```
OlEzhA0, line 6:
```suggestion
  const message = document.createElement('div');
```

OlEzhA0: check it for each your constant

src/scripts/main.js
```diff
@@ -1 +1,72 @@
...
+    resolve('Second promise was resolved');
+  });
+
+  BODY.addEventListener('contextmenu', () => {
+    resolve('Second promise was resolved');
+  });
+});
+
+const PROMISE3 = new Promise((resolve, reject) => {
+  let checker = '';
```
OlEzhA0, line 46: you can create 2 boolean variables like
`let leftClick = false`
`let rightClick = false`
and make it true when clicked

OlEzhA0 (changes requested):
hey! if you have any questions - ping me in fe chat

slavanosatskyi (approved):

PR #399, opened 2023-02-16
result

novak-k:
- [DEMO LINK](https://novak-k.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,67 @@
 'use strict';
+
+const body = document.querySelector('body');
+
+new Promise((resolve, reject) => {
+  body.addEventListener('click', () => {
+    resolve('First promise was resolved');
+  });
+
+  setTimeout(() => {
+    reject(new Error('First promise was rejected'));
+  }, 3000);
+}).then(data => {
+  successHandler(data);
+}).catch(error => {
+  errorHandler(error);
+});
+
+new Promise((resolve, reject) => {
+  body.addEventListener('click', () => {
+    resolve('Second promise was resolved');
+  });
+
+  body.addEventListener('contextmenu', () => {
+    resolve('Second promise was resolved');
+  });
+}).then(data => successHandler(data));
+
+new Promise((resolve, reject) => {
+  // eslint-disable-next-line no-shadow
+  const p1 = new Promise((resolve, reject) => {
+    body.addEventListener('click', () => {
+      resolve('left click');
+    });
+  });
+
+  // eslint-disable-next-line no-shadow
+  const p2 = new Promise((resolve, reject) => {
+    body.addEventListener('contextmenu', () => {
+      resolve('right click');
+    });
+  });
+
+  Promise.all([p1, p2]).then(data => {
+    resolve('Third promise was resolved');
+  });
+}).then(data => successHandler(data));
+
+const successHandler = (message) => {
+  const div = document.createElement('div');
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1 +1,67 @@
...
+      resolve('left click');
+    });
+  });
+
+  // eslint-disable-next-line no-shadow
+  const p2 = new Promise((resolve, reject) => {
+    body.addEventListener('contextmenu', () => {
+      resolve('right click');
+    });
+  });
```
AlisaVlasova, line 42: use 'mousedown' event here to manage left-right click

AlisaVlasova (changes requested):

src/scripts/main.js
novak-k: done,  as I understand it

src/scripts/main.js
```diff
@@ -1 +1,67 @@
...
+
+const errorHandler = (message) => {
+  const div = document.createElement('div');
+
+  div.setAttribute('data-qa', 'notification');
+  div.setAttribute('class', 'warning');
+  div.innerHTML = message;
+
+  body.append(div);
+};
```
sTorba24, line 67: I think you know where to move it :)

src/scripts/main.js
```diff
@@ -1 +1,67 @@
...
+  setTimeout(() => {
+    reject(new Error('First promise was rejected'));
+  }, 3000);
+}).then(data => {
+  successHandler(data);
+}).catch(error => {
+  errorHandler(error);
+});
+
+new Promise((resolve, reject) => {
```
sTorba24, line 19:
```suggestion
new Promise((resolve) => {
```

sTorba24: You can remove it if you are not using `reject` :)

src/scripts/main.js
```diff
@@ -1 +1,67 @@
...
+  body.addEventListener('click', () => {
+    resolve('First promise was resolved');
+  });
+
+  setTimeout(() => {
+    reject(new Error('First promise was rejected'));
+  }, 3000);
+}).then(data => {
```
sTorba24, line 13: move it to const and after add then and catch 
(with variable you can reuse if needed )

src/scripts/main.js
```diff
@@ -1 +1,67 @@
...
+      if (e.which === 3) {
+        resolve('right click');
+      }
+    });
+  });
+
+  Promise.all([p1, p2]).then(data => {
+    resolve('Third promise was resolved');
+  });
+}).then(data => successHandler(data));
```
sTorba24, line 47: you don't need to create a promise inside promise
try to create boolean variables that will kepp was clicked right and left buttons 
If clicked both => do action

sTorba24 (changes requested):
Nice work! Some things to improve :)
Also fix a demo link please

src/scripts/main.js
novak-k: sure +

src/scripts/main.js
novak-k: done!

src/scripts/main.js
novak-k: done

src/scripts/main.js
novak-k: okay, done

sTorba24 (approved):
Good for me, approved! :)

PR #409, opened 2023-02-24
Js promises practice task was made

SamirMamedov1010:
- [DEMO LINK](https://SamirMamedov1010.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,72 @@
 'use strict';
+
+const body = document.querySelector('body');
+
+const resolver1 = (resolve, reject) => {
+  document.addEventListener('mousedown', (e) => {
+    if (e.button === 0) {
+      resolve();
+    }
+  });
+
+  setTimeout(() => {
+    reject();
+  }, 3000);
+};
+
+const resolver2 = (resolve) => {
+  document.addEventListener('mousedown', (e) => {
+    if (e.button === 0 || e.button === 2) {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1 +1,72 @@
...
+
+const resolver3 = (resolve) => {
+  document.addEventListener('mousedown', (e1) => {
+    document.addEventListener('mousedown', (e2) => {
+      if ((e1.button === 0 && e2.button === 2)
+      || (e2.button === 0 && e1.button === 2)) {
+        resolve();
+      }
+    });
+  });
```
loralevitska, line 33: Don't use nested listener. Create 2 variables instead of and change it values after right and left click. So when you variables show you that there were clicks on both buttons resolve third promise

src/scripts/main.js
```diff
@@ -1 +1,72 @@
 'use strict';
+
+const body = document.querySelector('body');
+
+const resolver1 = (resolve, reject) => {
```
loralevitska, line 5: It's not resolver because it can be rejected. Better use variable names from task -> `firstPromise ` `secondPromise ` `thirdPromise `

loralevitska (changes requested):
A little more and it will be completed

SerhiiKirik (approved):
LGTM 🔥 
Would be better if you disable oncontextmenu

PR #411, opened 2023-02-24
Solution

pushkarskiyrodion:
[DEMO LINK](https://pushkarskiyrodion.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,57 @@
 'use strict';
+
+function createMessage(message, type, posTop, posRight) {
+  const div = document.createElement('div');
+
+  div.innerHTML = `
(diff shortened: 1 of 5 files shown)


src/scripts/main.js
```diff
@@ -1 +1,67 @@
...
+
+const secondPromise = new Promise((resolve) => {
+  document.addEventListener('click', () => {
+    resolve('Second promise was resolved');
+  });
+
+  document.addEventListener('contextmenu', (e) => {
+    e.preventDefault();
+    resolve('Second promise was resolved');
+  });
```
loralevitska, line 37: I guess in this case better use mousedown and check `e.button` instead of 2 listeners

loralevitska (approved):

PR #414, opened 2023-02-26
Promise practice

Igor-Yaremchuk:
[Promise practice](https://Igor-Yaremchuk.github.io/js_promises_practice_DOM/)


src/scripts/main.js
```diff
@@ -1 +1,73 @@
...
+    </div>
+  `);
+};
+
+const errorHandler = data => {
+  body.insertAdjacentHTML('beforeend', `
+    <div class="warning" data-qa="notification">
+      ${data}
+    </div>
+  `);
```
Anastasiia-Svintsova, line 60: You do almost the same here. Also could combine these methods to not repeat the logic. Just pass a class to the arguments

src/scripts/main.js
```diff
@@ -1 +1,73 @@
...
+      resolve('Third promise was resolved');
+    };
+  });
+
+  document.addEventListener('contextmenu', (e) => {
+    rightClick = true;
+
+    if (leftClick === true && rightClick === true) {
+      resolve('Third promise was resolved');
+    };
```
Anastasiia-Svintsova, line 43: Same. Could create a method for this part to not repeat yourself 
```
if (leftClick === true && rightClick === true) {
   resolve('Third promise was resolved');
};

Anastasiia-Svintsova (approved):

PR #415, opened 2023-02-27
Solution for Promises Practice Dom

ctk12:
[DEMO LINK](https://ctk12.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,71 @@
 'use strict';
+
+const body = document.querySelector('body');
+
+const notification = (message, type) => {
+  const div = document.createElement('div');
+
+  div.classList.add('notification', type);
+  div.innerText = message;
+  div.dataset.qa = 'notification';
+
+  body.append(div);
+};
+
+const messageS1 = () => {
+  notification('First promise was resolved', 'success');
+};
+
+const messageS2 = () => {
+  notification('Second promise was resolved', 'success');
+};
+
+const messageS3 = () => {
+  notification('Third promise was resolved', 'success');
+};
+
+const messageE1 = () => {
+  notification('First promise was rejected', 'warning');
+};
+
+const promise1 = new Promise((resolve, reject) => {
+  body.addEventListener('click', () => {
+    resolve();
+  });
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1 +1,71 @@
 'use strict';
+
+const body = document.querySelector('body');
+
+const notification = (message, type) => {
```
qoa11a, line 5: The name of the function should have information about what this function does. For example, this function could be named "createNotification".

src/scripts/main.js
```diff
@@ -1 +1,71 @@
...
+  notification('Second promise was resolved', 'success');
+};
+
+const messageS3 = () => {
+  notification('Third promise was resolved', 'success');
+};
+
+const messageE1 = () => {
+  notification('First promise was rejected', 'warning');
+};
```
qoa11a, line 29: I figure out that "S" means "success" and "E" means "error", but you should not name functions in that way. There's no need to create these functions as you don't re-use them, you can call "notification" where you need to.

src/scripts/main.js
```diff
@@ -1 +1,71 @@
...
+
+const promise1 = new Promise((resolve, reject) => {
+  body.addEventListener('click', () => {
+    resolve();
+  });
+
+  setTimeout(() => reject(messageE1), 3000);
```
qoa11a, line 36: Why do you pass messageE1 to reject? In your case, it's senseless.

qoa11a (changes requested):
Nicely done! 👍 
Few small fixes.

src/scripts/main.js
```diff
@@ -1 +1,55 @@
...
+
+  body.append(div);
+};
+
+const promise1 = new Promise((resolve, reject) => {
+  body.addEventListener('click', () => {
+    resolve();
+  });
+
+  setTimeout(() => reject(Error), 3000);
```
qoa11a, line 20:
```suggestion
  setTimeout(() => reject(new Error()), 3000);
```

qoa11a (approved):

PR #426, opened 2023-03-07
Chupiak

Chupiak:
[DEMO LINK](https://chupiak.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,70 @@
 'use strict';
+
+const firstPromiseResolvedMessage = 'First promise was resolved';
+const secondPromiseResolvedMessage = 'Second promise was resolved';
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1 +1,62 @@
...
+
+const thirdPromise = new Promise(resolve => {
+  Promise.all([leftClick, rightClick]).then(() => {
+    resolve(thirdResolveMessage);
+  });
+});
+
+firstPromise.then(success).catch(error);
+secondPromise.then(success);
+thirdPromise.then(success);
```
loralevitska, line 62: Use more meaningful naming for function. Now it's not obvious what is it. And if we'll find it in code it's not clear what the function does

loralevitska (changes requested):
Almost perfect)
Let's improve

PR #435, opened 2023-03-11
Promises practice

Oleksandr-Maslo:
https://oleksandr-maslo.github.io/js_promises_practice_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,53 @@
 'use strict';
+
+function insertMessage(messageClass, message) {
+  document.body.insertAdjacentHTML('beforeend',
+    `<div data-qa="notification" class = ${messageClass}> ${message}</div>`
+  );
+};
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1 +1,53 @@
 'use strict';
+
+function insertMessage(messageClass, message) {
+  document.body.insertAdjacentHTML('beforeend',
+    `<div data-qa="notification" class = ${messageClass}> ${message}</div>`
+  );
+};
+
+const promise1 = new Promise((resolve, reject) => {
+  document.addEventListener('click', () => resolve());
```
TarasHoliuk, line 10:
```suggestion
  document.addEventListener('click', () => resolve('First promise was resolved'));
```
just want to make sure you know it:
you can pass this message here and then get it as an argument of the callback function you passed in `then` method.
The same with error and `catch`

TarasHoliuk (approved):
Approved :heavy_check_mark:

PR #442, opened 2023-03-14
Solution

GUSILLUS:
- [DEMO LINK](https://GUSILLUS.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,70 @@
 'use strict';
+
+const firstPromise = new Promise((resolve, reject) => {
+  document.addEventListener('click', () => {
+    resolve('First promise was resolved');
+  });
+
+  setTimeout(reject, 3000);
+});
+
+firstPromise
+  .then(handleResolve)
+  .catch(handleReject);
+
+const secondPromise = new Promise((resolve, reject) => {
+  document.addEventListener('click', () => {
+    resolve('Second promise was resolved');
+  });
+
+  document.addEventListener('contextmenu', () => {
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1 +1,70 @@
...
+function handleResolve(result) {
+  const success = document.createElement('div');
+
+  success.setAttribute('data-qa', 'notification');
+  success.classList.add('success');
+  success.textContent = result;
+  document.body.append(success);
+}
+
+function handleReject() {
```
Viktor-Kost, line 63: Try not repeat yourself. This function does the same as `handleResolve`. 
So you can create a common function like `showMessage` and pass different arguments

src/scripts/main.js
```diff
@@ -1 +1,70 @@
...
+
+const secondPromise = new Promise((resolve, reject) => {
+  document.addEventListener('click', () => {
+    resolve('Second promise was resolved');
+  });
+
+  document.addEventListener('contextmenu', () => {
+    event.preventDefault();
+    resolve('Second promise was resolved');
+  });
```
Viktor-Kost, line 23: you have the same resolvers, so create own function at this scope to not repeat yourself. Apply for similar cases

```
const handleResolve() {
      resolve('Second promise was resolved');
    }
```
    
```suggestion
  document.addEventListener('contextmenu', (e) => {
    event.preventDefault();
    handleResolve();
  });
```

Viktor-Kost (changes requested):
Nice!

Some moments can be improved

PR #451, opened 2023-03-22
Solution

DaveBeetle:
[DEMO LINK](https://DaveBeetle.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -34,7 +34,7 @@ const promise3 = new Promise((resolve) => {
       right = true;
     }
 
-    if (left && right) {
+    if (left === true && right === true) {
       resolve('Third promise was resolved');
     }
   });


src/scripts/main.js
```diff
@@ -1 +1,74 @@
...
+    };
+  });
+
+  document.addEventListener('contextmenu', (e) => {
+    e.preventDefault();
+    right = true;
+
+    if (left === true && right === true) {
+      resolve('Third promise was resolved');
+    };
```
loralevitska, line 42: It can be simplify with 1 listener [`mousedown ` event](https://developer.mozilla.org/en-US/docs/Web/API/Element/mousedown_event) then check `event.button`

loralevitska (changes requested):
Almost perfect) Let's improve

OleziO (changes requested):
The third promise does not work.

nazarmatsevych:
> everything works fine on my laptop, i don't know why it doesn't work here 

Looks like e.button === 2 doesn't work on windows but work on ios

src/scripts/main.js
```diff
@@ -1 +1,71 @@
...
+  document.addEventListener('mousedown', e => {
+    if (e.button === 0) {
+      left = true;
+    }
+
+    if (e.button === 2) {
+      right = true;
+    }
+
+    if (left === true && right === true) {
```
nazarmatsevych, line 37:
```suggestion
    if (left && right) {
```

same but shorter)

nazarmatsevych (approved):

PR #469, opened 2023-04-07
Solution

Zibi95:


src/scripts/main.js
```diff
@@ -1 +1,59 @@
...
+  body.addEventListener('mousedown', ({ button }) => {
+    if (button === 0 || button === 2) {
+      resolve('Second promise was resolved');
+    }
+  });
+});
+
+secondPromise
+  .then(result => createDiv('success', result))
+  .catch(errorMsg => createDiv('error', errorMsg));
```
Radoslaw-Czerniawski, line 38: According to the requirements elements should have class `warning` rather than `error`

Radoslaw-Czerniawski (approved):
Small error (wrong class name), apart from that the implementation is very good, approved :)

PR #479, opened 2023-04-15
Completed task

metja:
- [DEMO LINK](https://metja.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,79 @@
 'use strict';
+
+const body = document.querySelector('body');
+
+const firsPromise = new Promise((resolve, reject) => {
+  const firstClick = body.addEventListener('click', () => {
+    resolve();
+  });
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1 +1,79 @@
...
+const firsPromise = new Promise((resolve, reject) => {
+  const firstClick = body.addEventListener('click', () => {
+    resolve();
+  });
+
+  if (!firstClick) {
+    setTimeout(() => {
+      reject(Error);
+    }, 3000);
+  };
```
SanyaBratashchuk, line 14:
```suggestion
    setTimeout(() => {
      reject(Error);
    }, 3000);
```
There is no sense in this if since addEventListener always returns undefined which is falsy

SanyaBratashchuk (approved):

PR #482, opened 2023-04-17
Solution

Vaskovics:
<a href="https://vaskovics.github.io/js_promises_practice_DOM/">demo</a>

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,85 @@
 'use strict';
+
+const successHandler = (message) => {
+  const newDiv = document.createElement('div');
+
+  newDiv.dataset.qa = 'notification';
+  newDiv.classList.add('success');
+
+  newDiv.innerText = message;
+
+  document.body.append(newDiv);
+};
+
+const errorHandler = (message) => {
+  const newDiv = document.createElement('div');
+
+  newDiv.dataset.qa = 'notification';
+  newDiv.classList.add('warning');
+
+  newDiv.innerText = message;
+
+  document.body.append(newDiv);
+};
+
+const firstPromise = new Promise((resolve, reject) => {
+  let eventOccurred = false;
+
+  document.addEventListener('click', () => {
+    const resolvedMessage = 'First promise was resolved!';
+
+    eventOccurred = true;
+
+    resolve(resolvedMessage);
+  });
+
+  if (!eventOccurred) {
+    setTimeout(() => {
+      const rejectedMessage = 'First promise was rejected';
+
+      reject(rejectedMessage);
+    }, 3000);
+  };
+});
+
+firstPromise
+  .then(successHandler)
+  .catch(errorHandler);
+
+const secondPromise = new Promise((resolve) => {
+  function clickHandler(newEvent) {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1 +1,85 @@
...
+const thirdPromise = new Promise((resolve) => {
+  let rightClick = false;
+  let leftClick = false;
+
+  document.addEventListener('mousedown', (myEvent) => {
+    myEvent.preventDefault();
+
+    if (myEvent.button === 0) {
+      leftClick = true;
+    }
```
anstsot, line 62: you can simplify to
```suggestion
    leftClick = myEvent.button === 0;
```

src/scripts/main.js
```diff
@@ -1 +1,85 @@
...
+const errorHandler = (message) => {
+  const newDiv = document.createElement('div');
+
+  newDiv.dataset.qa = 'notification';
+  newDiv.classList.add('warning');
+
+  newDiv.innerText = message;
+
+  document.body.append(newDiv);
+};
```
anstsot, line 23: you can create one function but with additional `type` param

anstsot (changes requested):

src/scripts/main.js
```diff
@@ -1 +1,83 @@
...
+
+// const errorHandler = (message) => {
+//   const newDiv = document.createElement('div');
+
+//   newDiv.dataset.qa = 'notification';
+//   newDiv.classList.add('warning');
+
+//   newDiv.innerText = message;
+
+//   document.body.append(newDiv);
```
IvanFesenko, line 22: remove commented code

src/scripts/main.js
```diff
@@ -1 +1,83 @@
...
+const errorHandler = (message, type) => {
+  const newDiv = document.createElement('div');
+
+  newDiv.dataset.qa = 'notification';
+  newDiv.classList.add(type);
+
+  newDiv.innerText = message;
+
+  document.body.append(newDiv);
+};
```
IvanFesenko, line 12: Fix naming, because it wotks with success too as well.

src/scripts/main.js
IvanFesenko: Not fixed

IvanFesenko (changes requested):

Vaskovics:
in this case it won't work, as I need to change just one of the rightClick or leftClick, in case when I will use ternary operator as you suggested in every occurred event it will change both of variables, so I never get both true. Thank you!

src/scripts/main.js
```diff
@@ -1 +1,72 @@
...
+  let leftClick = false;
+
+  document.addEventListener('mousedown', (myEvent) => {
+    if (myEvent.button === 0) {
+      leftClick = true;
+    }
+
+    if (myEvent.button === 2) {
+      rightClick = true;
+    }
```
IvanFesenko, line 61: Means that one
```suggestion
    leftClick = myEvent.button === 0;
    rightClick = myEvent.button === 2;
```

IvanFesenko (approved):

PR #483, opened 2023-04-17
Solution

dianaavem:
[DEMO LINK](https://dianaavem.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,60 @@
 'use strict';
+
+const html = document.querySelector('html');
+const body = document.querySelector('body');
+
+function createMessage(type, message) {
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1 +1,60 @@
...
+
+const promise2 = new Promise(resolve => {
+  html.addEventListener('click', () => {
+    resolve();
+  });
+
+  html.addEventListener('contextmenu', (e) => {
+    e.preventDefault();
+    resolve();
+  });
```
loralevitska, line 32: Can be simplify with one listener and [`event.button`](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button) checker

loralevitska (approved):

PR #488, opened 2023-04-19
promises were added

Pavlo-Petrashevskyi:


src/scripts/main.js
```diff
@@ -1 +1,67 @@
...
+
+    if (leftButton && rightButton) {
+      resolve('Third promise was resolved');
+    }
+  });
+});
+
+promise1
+  .then(success => createMessage(success, 'success'))
+  .catch(Err => createMessage(Err, 'error'));
```
DarkMistyRoom, line 59:
```suggestion
  .catch(Err => createMessage(Err, 'warning'));
```

DarkMistyRoom: fix it everywhere

DarkMistyRoom: image

src/scripts/main.js
```diff
@@ -1 +1,67 @@
...
+  if (messageClass === 'error') {
+    newMess = 'First promise was rejected';
+  };
+
+  document.body.insertAdjacentHTML('beforeend', `
+    <div data-qa="notification" class="${messageClass}">
+      ${newMess}
+    </div>
+  `);
+}
```
DarkMistyRoom, line 15: make this function flexible. It should work properly if we will add reject handlers for other promises

src/scripts/main.js
```diff
@@ -1 +1,67 @@
...
+  });
+
+  setTimeout(() => reject(Error('First promise was rejected')), 3000);
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  document.addEventListener('mousedown', () => {
+    resolve('Second promise was resolved');
+  });
+});
```
DarkMistyRoom, line 31: this promise resolves even on clicking additional mouse buttons(wheel). Make it work only for left and right clicks

src/scripts/main.js
```diff
@@ -1 +1,67 @@
...
+  document.addEventListener('mousedown', (e) => {
+    if (e.button === 0) {
+      leftButton = true;
+    };
+
+    if (leftButton && rightButton) {
+      resolve('Third promise was resolved');
+    }
+  });
+});
```
DarkMistyRoom, line 55: looks not really pretty: we have additional variables and duplication here. 

I suggest you change the approach and use two additional promises: leftClick and rightClick. Then you can manage these 3 promises based on them

src/scripts/main.js
```diff
@@ -1 +1,67 @@
...
+  .then(success => createMessage(success, 'success'))
+  .catch(Err => createMessage(Err, 'error'));
+
+promise2
+  .then(success => createMessage(success, 'success'))
+  .catch(Err => createMessage(Err, 'error'));
+
+promise3
+  .then(success => createMessage(success, 'success'))
+  .catch(Err => createMessage(Err, 'error'));
```
DarkMistyRoom, line 67: these two never reject

DarkMistyRoom (changes requested):

src/scripts/main.js
Pavlo-Petrashevskyi: Done

src/scripts/main.js
Pavlo-Petrashevskyi: Done

src/scripts/main.js
Pavlo-Petrashevskyi: Done

src/scripts/main.js
Pavlo-Petrashevskyi: I guess done

loralevitska (approved):

src/scripts/main.js
Pavlo-Petrashevskyi: Done

PR #496, opened 2023-04-25
Solution

dudar-y:
[DEMO LINK](https://dudar-y.github.io/js_promises_practice_DOM/)


tetiana-ivanchenko (changes requested):
if I click first the right mouse button, then the left mouse button, the third promise should resolve, but it doesn't

loralevitska (approved):
GJ

PR #501, opened 2023-04-27
solution

mariuszgit:


src/scripts/main.js
```diff
@@ -1 +1,61 @@
...
+
+    setTimeout(() => {
+      reject(new Error('First promise was rejected'));
+    }, 3000);
+  });
+}
+
+function secondPromise() {
+  return new Promise(resolve => {
+    document.addEventListener('mousedown', (e) => {
```
DorotaLeniecDev, line 18: This mousedown event can be triggered not only by left/right button on the mouse. You can add some condition to detect propper click. Check https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons

DorotaLeniecDev (changes requested):
Good job, one comment from my side :)

src/scripts/main.js
```diff
@@ -1 +1,63 @@
...
+      switch (e.button) {
+        case 0:
+          leftBtn = true;
+          break;
+        case 2:
+          rightBtn = true;
+          break;
+        default:
+          throw new Error('Click left or right button');
```
Radoslaw-Czerniawski, line 40: Don't throw unnecessary errors, especially when they are not handled.  Either exclude the default case from switch or just do a regular if statement with || operator

Radoslaw-Czerniawski (changes requested):
Very nice, needs one small adjustment

Radoslaw-Czerniawski (approved):
All good :D

PR #503, opened 2023-04-28
solution

NadieinOleh:
https://NadieinOleh.github.io/js_promises_practice_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,49 @@
 'use strict';
+
+const firstPromise = new Promise((resolve, reject) => {
+  document.body.addEventListener('click', () => {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1 +1,49 @@
...
+  document.body.addEventListener('click', () => {
+    resolve('Second promise was resolved');
+  });
+});
+
+const thirdPromise = new Promise((resolve, reject) => {
+  resolve('Third promise was resolved');
+});
+
+secondPromise
```
loralevitska, line 26: I guess here should be first and third promises too

Anastasiia-Svintsova (approved):

PR #504, opened 2023-05-01
Initial commit

RomanHrytsiuk:
https://RomanHrytsiuk.github.io/js_promises_practice_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,63 @@
 'use strict';
+
+const body = document.querySelector('body');
+const div = document.createElement('div');
+
+div.dataset.qa = 'notification';
+
+let firstClick = false;
+let secondClick = false;
+
+const firstPromise = new Promise((resolve, reject) => {
+  const delay = setTimeout(() => {
+    div.classList.add('warning');
+    div.innerText = 'First promise was rejected';
+    body.append(div);
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1 +1,63 @@
...
+div.dataset.qa = 'notification';
+
+let firstClick = false;
+let secondClick = false;
+
+const firstPromise = new Promise((resolve, reject) => {
+  const delay = setTimeout(() => {
+    div.classList.add('warning');
+    div.innerText = 'First promise was rejected';
+    body.append(div);
```
anstsot, line 15: create a separate function for adding the notification and reuse it everywhere. please create the `div` in that function each time you call it.

anstsot (changes requested):

src/scripts/main.js
```diff
@@ -1 +1,55 @@
...
+      resolve(addNotification('Third promise was resolved', 'success'));
+    }
+  });
+});
+
+firstPromise
+  .then()
+  .catch();
+
+secondPromise.then();
```
etojeDenys, line 53: You should call `addNotification` in then or catch

etojeDenys (changes requested):

SanyaBratashchuk (approved):

PR #507, opened 2023-05-05
Promises practise solution

annperetiatko:
[DEMO LINK]
https://annperetiatko.github.io/js_promises_practice_DOM/


src/scripts/main.js
```diff
@@ -1 +1,83 @@
 'use strict';
+
+const firstPromise = new Promise((resolve, reject) => {
+  const click = document.addEventListener('click', () => {
```
Volodymyr-Melnyk-Fron10der, line 4: click is always true, create a separated variable and set the true value only on click event

src/scripts/main.js
```diff
@@ -1 +1,83 @@
...
+    resolve();
+  });
+
+  if (!click) {
+    setTimeout((error) => {
+      reject(error);
+    }, 3000);
+  }
```
Volodymyr-Melnyk-Fron10der, line 12: you should start the timer anyway, move the condition into the timeout callback

src/scripts/main.js
```diff
@@ -1 +1,83 @@
...
+});
+
+firstPromise
+  .then(() => {
+    const div = document.createElement('div');
+
+    div.dataset.qa = 'notification';
+    div.className = 'success';
+    div.textContent = 'First promise was resolved';
+    document.body.appendChild(div);
```
Volodymyr-Melnyk-Fron10der, line 22: move this piece of code to the function and reuse it

src/scripts/main.js
```diff
@@ -1 +1,83 @@
...
+      const checkClicks = (leftClickNew, rightClickNew) => {
+        return new Promise((resolve) => {
+          if (leftClickNew && rightClickNew) {
+            resolve();
+          }
+        });
+      };
+
+      return checkClicks;
+    });
```
Volodymyr-Melnyk-Fron10der, line 73: just add 2 event listeners in the promise and 2 flags to check were they clicked already

Volodymyr-Melnyk-Fron10der (changes requested):

src/scripts/main.js
```diff
@@ -1 +1,76 @@
...
+  const div = document.createElement('div');
+
+  div.dataset.qa = 'notification';
+  div.className = `${className}`;
+  div.textContent = `${promiseName} promise was ${condition}`;
+  document.body.appendChild(div);
+}
+
+const firstPromise = new Promise((resolve, reject) => {
+  let clicked = false;
```
ivanchuk-yurii, line 13: there is no need in this variable, you can just `removeTimeout` on click

src/scripts/main.js
```diff
@@ -1 +1,76 @@
...
+
+const rightClick = new Promise((resolve) => {
+  document.addEventListener('contextmenu', () => {
+    rightClicked = true;
+    resolve();
+  });
+});
+
+const thirdPromise = new Promise((resolve) => {
+  Promise.all([leftClick, rightClick]).then(() => {
```
ivanchuk-yurii, line 66: `Promise.all` resolves only if all promises are resolved. so `leftClicked && rightClicked` is redundant

ivanchuk-yurii (changes requested):

ivanchuk-yurii (approved):

PR #510, opened 2023-05-07
need help

metinbicaksiz:
[DEMO LINK](https://metinbicaksiz.github.io/js_promises_practice_DOM/)

I have an issue with a few of the messages. Can you help me figure it out?


src/scripts/main.js
```diff
@@ -1 +1,66 @@
...
+    }
+  });
+});
+
+firstPromise.then((message) => {
+  notification.classList.add('success');
+  notification.textContent = message;
+}).catch((error) => {
+  notification.classList.add('warning');
+  notification.textContent = error.message;
```
polosanya, line 25: Create a separate function for showing a notification as the logic is almost the same for all promises

src/scripts/main.js
```diff
@@ -1 +1,66 @@
...
+});
+
+const secondPromise = new Promise((resolve) => {
+  function handleClick(e) {
+    if (e.button === 0 || e.button === 2) {
+      document.removeEventListener('click', handleClick);
+      resolve('Second promise was resolved');
+    }
+  }
+  document.addEventListener('click', handleClick);
```
polosanya, line 35: Try contextmenu event instead

src/scripts/main.js
```diff
@@ -1 +1,66 @@
...
+
+secondPromise.then((message) => {
+  notification.classList.add('success');
+  notification.textContent = message;
+});
+
+const thirdPromise = new Promise((resolve) => {
+  const clicks = {
+    left: false, right: false,
+  };
```
polosanya, line 46: Let's try to move them to the upper scope and change for other eventlisteners too

polosanya (changes requested):
If you have any questions - it's better to ask them in js_chat 🙂

src/scripts/main.js
```diff
@@ -1 +1,65 @@
 'use strict';
+
+const notification = document.querySelector('[data-qa="notification"]');
+const clicks = {
+  left: false,
+  right: false,
+};
+
+const messageFunction = (message) => {
```
polosanya, line 9:
```suggestion
const showMessage = (message) => {
```

src/scripts/main.js
```diff
@@ -1 +1,65 @@
...
+  notification.classList.add('success');
+  notification.textContent = message;
```
polosanya, line 11: You should create a new notification for each message

src/scripts/main.js
```diff
@@ -1 +1,65 @@
...
+
+secondPromise.then(messageFunction);
+
+const thirdPromise = new Promise((resolve) => {
+  function handleClick(e) {
+    if (e.button === 0) {
+      clicks.left = true;
+    } else if (e.button === 2) {
+      clicks.right = true;
+    }
```
polosanya, line 55: You can change clicks in other promises too

src/scripts/main.js
```diff
@@ -1 +1,65 @@
...
+
+    if (clicks.left && clicks.right) {
+      document.addEventListener('click', handleClick);
+      resolve('Third promise was resolved');
+    }
+  }
+  document.addEventListener('click', handleClick);
```
polosanya, line 62: Why do you need to have too similar eventListeners? Also, better check left and right click inside eventListener and resolve the promise if both of them are true

polosanya (changes requested):
Let's try to make it working

PR #515, opened 2023-05-12
FedenkoPromisesPractice

VladFedenko:
https://VladFedenko.github.io/js_promises_practice_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,78 @@
 'use strict';
+
+const body = document.querySelector('body');
+const message = document.createElement('div');
+
+function createPromise() {
+  const resolver = (resolve, reject) => {
+    body.addEventListener('click', () => {
+      resolve('First promise was resolved');
+    });
+
+    setTimeout(() => {
+      reject('First promise was rejected');
+    }, 3000);
+  };
+
+  return new Promise(resolver);
+}
+
+function createPromise2() {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1 +1,78 @@
...
+  .then(() => {
+    message.innerHTML = 'Third promise was resolved';
+    message.classList = 'warning';
+    body.append(message);
+  })
+  .catch(result => {
+    message.innerHTML = result;
+    message.classList = 'warning';
+    body.append(message);
+  });
```
IvanFesenko, line 78: All 3 promises should be independed by each other. 

```
promise1.then().catch()
promise2.then().catch()
promise3.then().catch()
```

Also I should see 3 different messages. Something like this
2023-05-15_14h49_09
2023-05-15_14h49_22

IvanFesenko (changes requested):
Some refactor required

src/scripts/main.js
```diff
@@ -1 +1,70 @@
...
+}
+
+const promise1 = createPromise();
+const promise2 = createPromise2();
+const promise3 = createPromise3();
+
+promise1.then(result => {
+  message.innerHTML += result;
+})
```
mykhalenych, line 55:
```suggestion
}).catch(result => {
  message.innerHTML += result;
});
```

mykhalenych (approved):

PR #516, opened 2023-05-14
Solution

oksana-skuba:
[DEMO](https://oksana-skuba.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,68 @@
 'use strict';
+
+function createMessage(className, message) {
+  const div = document.createElement('div');
+
+  div.dataset.qa = 'notification';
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1 +1,68 @@
...
+    }
+  });
+});
+
+firstPromise
+  .then(result => {
+    createMessage('success', result);
+  })
+  .catch(result => {
+    createMessage('warning', result);
```
marina-bidenko, line 57: Change the catch block parameter name from 'result' to 'error', as it's more appropriate for a rejected promise.

marina-bidenko (approved):
Good job

PR #518, opened 2023-05-14
Solution

Ihor-Zakharov:
[DEMO LINK](https://ihor-zakharov.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,77 @@
-'use strict';
+"use strict";
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1 +1,77 @@
-'use strict';
+"use strict";
+
+document.addEventListener('contextmenu', (e) => {
+  e.preventDefault();
+})
```
anstsot, line 5: why do you need this listener?

PR #520, opened 2023-05-15
solution

AnastasiiaVorobets:
- [DEMO LINK](https://AnastasiiaVorobets.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,57 @@
 'use strict';
+
+const firstPromise = new Promise((resolve, reject) => {
+  document.addEventListener('click', () => {
+    resolve('First promise was resolved');
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1 +1,57 @@
...
+    resolve('Second promise was resolved');
+  });
+});
+
+const thirdPromise = new Promise((resolve, reject) => {
+  document.addEventListener('click', () => {
+    document.addEventListener('contextmenu', () => {
+      resolve('Third promise was resolved');
+    });
+  });
```
vlad-demchuk, line 28: it will be more readable not to nest two listeners

vlad-demchuk (changes requested):
implement this point
image

PR #524, opened 2023-05-17
add task solution

NikaNika12:
https://NikaNika12.github.io/js_promises_practice_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,79 @@
 'use strict';
+
+const body = document.querySelector('body');
+const message = document.createElement('div');
+
+const firstPromise = new Promise((resolve, reject) => {
+  body.addEventListener('click', () => {
+    resolve('First promise was resolved');
+  });
+
+  setTimeout(() => {
+    reject(new Error('First promise was rejected'));
+  }, 3000);
+});
+
+firstPromise
+  .then(result => {
+    message.setAttribute('data-qa', 'notification');
+    message.className = 'success';
+    message.innerText = result;
+    body.append(message);
+  })
+  .catch(error => {
+    message.setAttribute('data-qa', 'notification');
+    message.className = 'warning';
+    message.innerText = error;
+    body.append(message);
+  });
+
+const secondPromise = new Promise((resolve) => {
+  body.addEventListener('mousedown', (e) => {
+    if (e.button === 0 || e.button === 2) {
+      resolve('Second promise was resolved');
+    }
+  });
+});
+
+secondPromise
+  .then(result => {
+    message.setAttribute('data-qa', 'notification');
+    message.className = 'success';
+    message.innerText = result;
+    body.append(message);
+  })
+  .catch(null);
+
+const thirdPromise = new Promise((resolve) => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1 +1,79 @@
...
+  body.addEventListener('click', () => {
+    leftBtn = true;
+
+    if (leftBtn && rightBtn) {
+      resolve('Third promise was resolved');
+    }
+    /* if (e.button === 0 && e.button === 2) {
+      leftBtn = true;
+      rightBtn = true;
+      resolve('Third promise was resolved');
```
AlisaVlasova, line 60: delete commented code

src/scripts/main.js
```diff
@@ -1 +1,79 @@
...
+    } */
+  });
+
+  body.addEventListener('contextmenu', () => {
```
AlisaVlasova, line 64: it ll be better to use mousedown event here and on left button

AlisaVlasova (changes requested):

src/scripts/main.js
```diff
@@ -1 +1,78 @@
...
+
+  body.addEventListener('mousedown', (e) => {
+    if (e.button === 2) {
+      rightBtn = true;
+
+      if (leftBtn && rightBtn) {
+        resolve('Third promise was resolved');
+      }
+    }
+  });
```
sTorba24, line 69: this logic can be in one listener

src/scripts/main.js
```diff
@@ -1 +1,78 @@
...
+    reject(new Error('First promise was rejected'));
+  }, 3000);
+});
+
+firstPromise
+  .then(result => {
+    message.setAttribute('data-qa', 'notification');
+    message.className = 'success';
+    message.innerText = result;
+    body.append(message);
```
sTorba24, line 21: create a function for adding message and reuse it everywhere you need

sTorba24 (changes requested):
Good work! Some things to improve

src/scripts/main.js
```diff
@@ -1 +1,68 @@
...
+  message.innerText = result;
+  body.append(message);
+}
+
+function handleError(error) {
+  message.setAttribute('data-qa', 'notification');
+  message.className = 'warning';
+  message.innerText = error;
+  body.append(message);
+}
```
anstsot, line 18: you can create one function for message adding with `type` param and reuse it. 
please move `message` div creation to this function because it rewrites the message every time

anstsot (changes requested):

src/scripts/main.js
NikaNika12: unction handler(type) {
  const message = document.createElement('div');

  message.setAttribute('data-qa', 'notification');

  if (type === 'result') {
    message.className = 'success';
    message.innerText = result;
  } else if (type === 'error') {
    message.className = 'warning';
    message.innerText = error;
  }

  body.append(message);
}

NikaNika12: а как это сделать у меня ошибки выводит тогда

OleziO (approved):

PR #526, opened 2023-05-18
promises practice DOM - first try

RufatRM:
https://RufatRM.github.io/js_promises_practice_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,74 @@
 'use strict';
+
+let leftClick = false;
+let rightClick = false;
+
+const succesDiv1 = document.createElement('div');
+const errorDiv1 = document.createElement('div');
+const succesDiv2 = document.createElement('div');
+const succesDiv3 = document.createElement('div');
+
+succesDiv1.classList.add('message');
+errorDiv1.classList.add('warning');
+succesDiv2.classList.add('message');
+succesDiv3.classList.add('message');
+
+succesDiv1.setAttribute('data-qa', 'notification');
+errorDiv1.setAttribute('data-qa', 'notification');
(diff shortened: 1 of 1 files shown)


Anastasiia-Svintsova (changes requested):
- This task is not implemented 
image

https://github.com/mate-academy/js_promises_practice_DOM/assets/99403198/13cc1f03-99c0-4b38-95c7-2f531612dcb0

src/scripts/main.js
```diff
@@ -1 +1,71 @@
...
+    if (leftClick && rightClick) {
+      resolve('Third promise was resolved');
+    }
+  });
+});
+
+function showMessage(message) {
+  const tag = document.createElement('div');
+
+  if (message.includes('resolved')) {
```
Anastasiia-Svintsova, line 40: This task is still not implemented. You can't use `includes` for the error and success. Because in the catch you receive an object as an argument

src/scripts/main.js
```diff
@@ -1 +1,71 @@
...
+
+  document.body.appendChild(tag);
+}
+
+promise1
+  .then((resolveMessage) => {
+    showMessage(resolveMessage);
+  })
+  .catch((errorMessage) => {
+    showMessage(errorMessage);
```
Anastasiia-Svintsova, line 60: I suppose it should work.
```suggestion
    showMessage(errorMessage.message);
```

Anastasiia-Svintsova (changes requested):

polosanya (changes requested):
Almost done! "First promise was rejected" message isn't shown after 3 seconds. 
Make sure that other promises work according to the task description too.
P.S. also, don't forget to redeploy your solution

qoa11a (approved):

PR #536, opened 2023-05-26
add task solution

RomanOstrous:
- [DEMO LINK](https://RomanOstrous.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,60 @@
 'use strict';
+
+const firstPromise = new Promise((resolve, reject) => {
+  document.addEventListener('click', () => {
+    resolve(`First promise was resolved`);
+  });
+
+  setTimeout(() => {
+    reject(Error('First promise was rejected'));
+  }, 3000);
+});
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1 +1,60 @@
...
+
+const contextmenu = new Promise((resolve) => {
+  document.addEventListener('contextmenu', () => {
+    resolve();
+  });
+});
+
+const thirdPromise = new Promise((resolve) => {
+  Promise.all([click, contextmenu])
+    .then(() => resolve('Third promise was resolved'));
```
polosanya, line 37: 🔥

src/scripts/main.js
```diff
@@ -1 +1,60 @@
...
+});
+
+const success = (suc) => {
```
polosanya, line 40:
```suggestion
const showSuccessMessage = (successMessage) => {
```
Better name functions according to these rules
image

polosanya (approved):
Nice!

PR #537, opened 2023-05-28
js_promises_practice_DOM_solution

YelyzavetaMolochynska:
[DEMO LINK](https://YelyzavetaMolochynska.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,64 @@
 'use strict';
+
+const body = document.querySelector('body');
+
+const successHandler = (message) => {
+  body.insertAdjacentHTML('beforebegin', `
+  <div data-qa="notification" class="success">${message}</div>
+  `);
+};
+
+const errorHandler = (message) => {
+  body.insertAdjacentHTML('beforebegin', `
+  <div data-qa="notification" class="warning">${message}</div>
+  `);
+};
+
+const firstPromise = new Promise((resolve, reject) => {
+  let clicked = false;
+
+  body.addEventListener('click', (e) => {
+    if (e.button === 0) {
+      clicked = true;
+      resolve('First promise was resolved');
+    }
+  });
+
+  setTimeout(() => {
+    if (!clicked) {
+      reject(new Error('First promise was rejected'));
+    }
+  }, 3000);
+});
+
+const secondPromise = new Promise((resolve) => {
+  document.body.addEventListener('click', (e) => {
+    resolve('Second promise was resolved');
+  });
+});
+
+const rightClick = new Promise((resolve) => {
+  body.addEventListener('click', (e) => {
+    resolve();
+  });
+});
+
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1 +1,64 @@
...
+    if (!clicked) {
+      reject(new Error('First promise was rejected'));
+    }
+  }, 3000);
+});
+
+const secondPromise = new Promise((resolve) => {
+  document.body.addEventListener('click', (e) => {
+    resolve('Second promise was resolved');
+  });
```
VitaliyBondarenko1982, line 37: for second promise also should be listener for right button
Screenshot from 2023-05-28 15-28-17

src/scripts/main.js
```diff
@@ -1 +1,64 @@
...
+  body.insertAdjacentHTML('beforebegin', `
+  <div data-qa="notification" class="success">${message}</div>
+  `);
+};
+
+const errorHandler = (message) => {
+  body.insertAdjacentHTML('beforebegin', `
+  <div data-qa="notification" class="warning">${message}</div>
+  `);
+};
```
VitaliyBondarenko1982, line 15: successHandler and errorHandler have same logic. it can be one function, which also accept className as argument

src/scripts/main.js
```diff
@@ -1 +1,64 @@
 'use strict';
+
+const body = document.querySelector('body');
+
+const successHandler = (message) => {
+  body.insertAdjacentHTML('beforebegin', `
```
VitaliyBondarenko1982, line 6:
```suggestion
  body.insertAdjacentHTML('beforeend', `
```

src/scripts/main.js
```diff
@@ -1 +1,64 @@
...
+  <div data-qa="notification" class="success">${message}</div>
+  `);
+};
+
+const errorHandler = (message) => {
+  body.insertAdjacentHTML('beforebegin', `
```
VitaliyBondarenko1982, line 12:
```suggestion
  body.insertAdjacentHTML('beforeend', `
```
notification content should be in body

VitaliyBondarenko1982 (changes requested):

src/scripts/main.js
```diff
@@ -1 +1,66 @@
...
+    }
+  }, 3000);
+});
+
+const secondPromise = new Promise((resolve) => {
+  document.body.addEventListener('click', (e) => {
+    if (e.button === 0 || e.button === 2) {
+      resolve('Second promise was resolved');
+    }
+  });
```
VitaliyBondarenko1982, line 33: not fixed
here should be listener also for 'contextmenu' event
something like that
```
   document.addEventListener('contextmenu', (e) => {
    e.preventDefault();
    resolve('Second promise was resolved');
  });

VitaliyBondarenko1982 (approved):

PR #546, opened 2023-06-04
promises practice

MariaShmakova2014:
[DEMO LINK](https://MariaShmakova2014.github.io/js_promises_practice_DOM/)

Can't understand:why I have undefined after the rejection of the first promise, why the second promise appears simultaneously with the first

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,56 @@
 'use strict';
+
+const body = document.body;
+
+function createNotification(message, classValue) {
+  const notification = document.createElement('div');
+
+  notification.setAttribute('data-qa', 'notification');
+  notification.innerText = message;
+  notification.classList = classValue;
+  body.append(notification);
+}
+
+const firstPromise = new Promise((resolve, reject) => {
+  body.addEventListener('click', () => {
+    resolve('First promise was resolved');
+  });
+
+  setTimeout(() => {
+    reject(Error('First promise was rejected'));
+  }, 3000);
+});
+
+function waitFor(evnt, number) {
+  return new Promise((resolve) => {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1 +1,56 @@
...
+  return new Promise((resolve) => {
+    document.addEventListener(evnt, () => {
+      resolve(`${number} promise was resolved`);
+    });
+  });
+}
+
+firstPromise
+  .catch(error => {
+    createNotification(error, 'warning');
```
etojeDenys, line 34: Change the order of 'catch' and 'then' to avoid skipping error handling for the firstPromise.

etojeDenys (changes requested):

Viktor-Kost (changes requested):
Hello!)
Third promise is resolved twice
Screenshot 2023-06-05 at 22 23 00

As a hint. You can use something like this
```
I would recommend to move your resolvers into own functions
const clickPromise = new Promise(resolve => {
  document.addEventListener('click', () => {
    resolve();
  });
});

const contextMenuPromise = new Promise(resolve => {
  document.addEventListener('contextmenu', () => {
    resolve();
  });
});
```

apply Promise.race to the secondPromise
```
const secondPromise = new Promise((resolve, reject) => {
  Promise.race([clickPromise, contextMenuPromise])
    .then((value) => resolve('Second promise was resolved'));
});
```

apply Promise.all to the thirdPromise
```
const thirdPromise = new Promise(resolve => {
  Promise.all([clickPromise, contextMenuPromise])
    .then((value) => resolve('Third promise was resolved'));
});
```

etojeDenys (approved):
lgtm

PR #552, opened 2023-06-15
Add click behawiors

AmadeuszAndroid:


src/scripts/main.js
```diff
@@ -1 +1,84 @@
...
+addHandler(secondPromise);
+addHandler(thirdPromise);
+
+const divSpaces = `.promiseMessage {
+  margin-left: 10px;
+}`;
+
+const style = document.createElement('style');
+
+style.innerText = divSpaces;
```
darokrk, line 83: think this should be rather `style.textContent`

darokrk (approved):
Approved 👍🏽  you can fix my one mention, but it's not necessary :)

src/scripts/main.js
AmadeuszAndroid: I changed to `textContent`.

PR #553, opened 2023-06-22
1.0.0 initial

domkab:
https://domkab.github.io/js_promises_practice_DOM

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,105 @@
 'use strict';
+
+const firstPromise = new Promise((resolve, reject) => {
+  document.addEventListener('click', e => {
+    resolve('First promise was resolved');
+  }, { once: true });
+
+  setTimeout(() => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1 +1,105 @@
...
+
+thirdPromise
+  .then((message) => {
+    const newDiv = document.createElement('DIV');
+
+    newDiv.setAttribute('data-qa', 'notification');
+    newDiv.classList.add('success');
+    newDiv.textContent = message;
+
+    document.body.appendChild(newDiv);
```
witflash, line 95: Could you move this logic (for every promises you used for success statement as well as for catch block) to the separate function-helper and use it everywhere you need?

witflash (changes requested):
Overall good work.
But let's make improvements. 
I've left you comment. Check it, please.

witflash (approved):
Great!
Well done 👍

PR #559, opened 2023-07-21
js_promises_practice_DOM

KarolinaWiktoria:
https://KarolinaWiktoria.github.io/js_promises_practice_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,73 @@
 'use strict';
+
+function createMessage(text, isError = false) {
+  const messageDiv = document.createElement('div');
+
+  messageDiv.dataset.qa = 'notification';
+
+  if (!isError) {
+    messageDiv.classList.add('success');
+  } else {
+    messageDiv.classList.add('warning');
+  }
+
+  messageDiv.textContent = text;
+  document.body.appendChild(messageDiv);
+}
+
+let documentClicked = false;
+
+const firstPromise = new Promise((resolve, reject) => {
+  document.addEventListener('click', function() {
+    resolve('First promise was resolved');
+
+    documentClicked = true;
+  });
+
+  if (!documentClicked) {
+    setTimeout(() => {
+      reject(new Error('First promise was rejected'));
+    }, 3000);
+  }
+});
+
+const secondPromise = new Promise((resolve) => {
+  document.addEventListener('mousedown', function() {
+    resolve('Second promise was resolved');
+  });
+});
+
+const thirdPromise = new Promise((resolve) => {
+  let rightClick = false;
+
+  document.addEventListener('click', function() {
+    if (rightClick) {
+      resolve('Third promise was resolved');
+    } else {
+      documentClicked = true;
+    }
+  });
+
+  document.addEventListener('contextmenu', function() {
+    if (documentClicked) {
+      resolve('Third promise was resolved');
+    } else {
+      rightClick = true;
+    }
+  });
+});
+
+firstPromise.then((success) => {
+  createMessage(success);
+})
+  .catch((error) => {
+    createMessage(error, true);
+  });
+
+secondPromise.then((success) => {
+  createMessage(success);
+});
+
+thirdPromise.then((success) => {
+  createMessage(success);
+});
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1 +1,73 @@
...
+});
+
+const secondPromise = new Promise((resolve) => {
+  document.addEventListener('mousedown', function() {
+    resolve('Second promise was resolved');
+  });
+});
+
+const thirdPromise = new Promise((resolve) => {
+  let rightClick = false;
```
DorotaLeniecDev, line 41: Add `let leftClick` declaration to this promise

DorotaLeniecDev (changes requested):
Check my comments

src/scripts/main.js
```diff
@@ -1 +1,73 @@
...
+    }
+  });
+});
+
+firstPromise.then((success) => {
+  createMessage(success);
+})
+  .catch((error) => {
+    createMessage(error, true);
+  });
```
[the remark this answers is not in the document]
KarolinaWiktoria: Thank you for all these suggestions. Fixed :)

src/scripts/main.js
```diff
@@ -1 +1,74 @@
...
+    }
+  });
+
+  document.addEventListener('contextmenu', function() {
+    if (documentClicked) {
+      resolve('Third promise was resolved');
+    } else {
+      rightClick = true;
+    }
+  });
```
DorotaLeniecDev, line 54: You could do the same in this event habdler
```suggestion
  document.addEventListener('contextmenu', function() {
      rightClick = true;
    
      if (rightClick && leftClick) {
        resolve(//with proper message);
      }
  });
```

src/scripts/main.js
```diff
@@ -1 +1,74 @@
...
+
+const thirdPromise = new Promise((resolve) => {
+  let rightClick = false;
+  let leftClick = true;
+
+  document.addEventListener('click', function() {
+    leftClick = true;
+
+    if (rightClick && leftClick) {
+      resolve();
```
DorotaLeniecDev, line 44: Add proper message

src/scripts/main.js
```diff
@@ -1 +1,74 @@
...
+    messageDiv.classList.add('success');
+  } else {
+    messageDiv.classList.add('warning');
+  }
+
+  messageDiv.textContent = text;
+  document.body.appendChild(messageDiv);
+}
+
+const documentClicked = false;
```
DorotaLeniecDev, line 18: This will no longer be needed
```suggestion

```

DorotaLeniecDev (changes requested):
Few more :)

src/scripts/main.js
```diff
@@ -1 +1,72 @@
...
+
+const secondPromise = new Promise((resolve) => {
+  document.addEventListener('mousedown', function() {
+    resolve('Second promise was resolved');
+  });
+});
+
+const thirdPromise = new Promise((resolve) => {
+  let rightClick = false;
+  let leftClick = true;
```
Radoslaw-Czerniawski, line 36: Initial value here should be false rather than true

src/scripts/main.js
```diff
@@ -1 +1,72 @@
...
+
+  document.addEventListener('click', function() {
```
Radoslaw-Czerniawski, line 38: This could be shortened, instead of registering two events, one on click and the other on the contextmenu you could register only one `mousedown` and based on the [button](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button) property on the event change leftClick and rightClick:
```js
body.addEventListener('mousedown', ({ button }) => {
  if (button === 0) {
    left = true;
  }

  if (button === 2) {
    right = true;
  }

  if (left && right) {
    resolve('Third promise was resolved');
  }
});
```

Radoslaw-Czerniawski (changes requested):
`leftClick` needs to be changed to initially be false, other comment is a suggestion (you may but don't need to)

src/scripts/main.js
KarolinaWiktoria: Oh, that's a nice solution. Thank you! :)

DorotaLeniecDev (approved):
Good job 💪

PR #561, opened 2023-08-01
solution

JOWISSA:
https://JOWISSA.github.io/js_promises_practice_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,75 @@
 'use strict';
+
+const notificationDiv = document.querySelector('[data-qa="notification"]');
+const body = document.querySelector('body');
+
+let leftClicked = false;
+let rightClicked = false;
+
+const firstPromise = new Promise((resolve, reject) => {
+  const clickHandler = (e) => {
+    if (e.button === 0) {
+      resolve('First promise was resolved on a left click in the document');
+    }
+  };
+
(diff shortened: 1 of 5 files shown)


src/scripts/main.js
```diff
@@ -1 +1,75 @@
...
+
+  setTimeout(() => {
+    body.removeEventListener('click', clickHandler);
+    reject(new Error('First promise was rejected in 3 seconds if not clicked'));
+  }, 3000);
+});
+
+firstPromise.then(
+  (message) => showNotification('success', message),
+  (error) => showNotification('warning', error.message)
```
TarasHoliuk, line 26: FYI: it works if you explicitly `reject` in the promise, but will not work if some error is thrown unexpectedly - for this purpose you need `catch`

src/scripts/main.js
```diff
@@ -1 +1,75 @@
...
+
+secondPromise.then((message) => showNotification('success', message));
+
+const thirdPromise = new Promise((resolve) => {
+  const leftClickHandler = () => {
+    leftClicked = true;
+    checkAndResolve();
+  };
+
+  const rightClickHandler = () => {
```
TarasHoliuk, line 47: prevent appearing context menu

TarasHoliuk (changes requested):

OleziO (approved):

PR #562, opened 2023-08-06
Solution

Patryk-Filipiak:
- [DEMO LINK](https://Patryk-Filipiak.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -45,12 +45,14 @@ const secondPromise = new Promise(resolve => {
 });
 
 const thirdPromise = new Promise((resolve) => {
-  document.addEventListener('mousedown', function eventListener({ button }) {
+  const mouseHandler = ({ button }) => {
     if ([0, 2].includes(button) && button !== lastButton) {
       resolve(THIRD_PROMISE.resolve);
-      document.removeEventListener('mousedown', eventListener);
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1 +1,63 @@
...
+    };
+  }, onlyOnce);
+});
+
+const thirdPromise = new Promise((resolve) => {
+  document.addEventListener('mousedown', function eventListener({ button }) {
+    if ([0, 2].includes(button) && button !== lastButton) {
+      resolve(THIRD_PROMISE.resolve);
+      document.removeEventListener('mousedown', eventListener);
+    };
```
choeqq, line 52: typically we use arrow functions for callbacks
```suggestion
  document.addEventListener('mousedown', ({ button }) => {
    if ([0, 2].includes(button) && button !== lastButton) {
      resolve(THIRD_PROMISE.resolve);
      document.removeEventListener('mousedown', eventListener);
    };
```

choeqq (changes requested):

src/scripts/main.js
Patryk-Filipiak: I updated to use the arrow function and exported the eventListener callback to const "handler", so before the update I specifically didn't use the arrow function to set the function name to easily remove the mouse listener as it was redundant already

PR #565, opened 2023-08-08
add task solution

ollavka:
[DEMO LINK](https://ollavka.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,67 @@
 'use strict';
+/* eslint-disable prefer-promise-reject-errors */
+
+const body = document.body;
+
+const appendNotification = (message, isError = false) => {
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1 +1,67 @@
...
+  const resolvePromise = () => resolve('Second promise was resolved');
+
+  document.addEventListener('click', resolvePromise);
+
+  document.addEventListener('contextmenu', resolvePromise);
+});
+
+const thirdPromise = new Promise((resolve) => {
+  let leftClicked = false;
+  let rightClicked = false;
```
DanilWeda, line 46: Can use an object instead let.

DanilWeda (approved):

PR #567, opened 2023-09-16
Add task solution

spojrzenie:


src/index.html
```diff
@@ -6,7 +6,7 @@
     <title>Promises practice</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
+  <body id="document">
```
darokrk, line 9: Avoid using IDs for styling. A better practice is to use classes instead.

src/scripts/main.js
```diff
@@ -1 +1,63 @@
 'use strict';
+
+const body = document.getElementById('document');
```
darokrk, line 3: There is no element with id 'document'. It's better to use document.body to select the body element.

darokrk (changes requested):
Looks good 👏🏽  small mention to fix and we good to go 🏆

maxim2310 (approved):
Good work!!!

PR #569, opened 2023-09-19
task done

tecHamster:
- [DEMO LINK](https://tecHamster.github.io/js_promises_practice_DOM/)


IvanFesenko (approved):
Good work, but too complicated and have redundant logic, that could be simplified

PR #575, opened 2023-10-21
solution

Duster-1:
https://duster-1.github.io/js_promises_practice_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,65 @@
 'use strict';
+
+const body = document.querySelector('body');
+
+function succesMessage(succesText) {
+  const newMessage = document.createElement('div');
+
+  newMessage.classList.add('succes');
+  newMessage.dataset.qa = 'notification';
+  newMessage.innerText = succesText;
+  body.append(newMessage);
+}
+
+function errorMessage(errorText) {
+  const newMessage = document.createElement('div');
+
+  newMessage.classList.add('warning');
+  newMessage.dataset.qa = 'notification';
+  newMessage.innerText = errorText;
+  body.append(newMessage);
+}
+
+const promise1 = new Promise((resolve, reject) => {
+  body.addEventListener('contextmenu', (e) => {
+    e.preventDefault();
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1 +1,65 @@
...
+}
+
+function errorMessage(errorText) {
+  const newMessage = document.createElement('div');
+
+  newMessage.classList.add('warning');
+  newMessage.dataset.qa = 'notification';
+  newMessage.innerText = errorText;
+  body.append(newMessage);
+}
```
etojeDenys, line 21: actually, you can combine it in one function, and just pass the class as the second argument

etojeDenys (changes requested):

src/scripts/main.js
```diff
@@ -1 +1,84 @@
...
+  let isRightClicked = false;
+
+  body.addEventListener('mousedown', (e) => {
+    if (e.button === 0) {
+      isLeftClicked = true;
+    }
+
+    if (e.button === 2) {
+      isRightClicked = true;
+    }
```
VitaliyBodnar, line 63: it's better to use switch case statement here instead of multiple if blocks

src/scripts/main.js
```diff
@@ -1 +1,84 @@
...
+const promise1 = new Promise((resolve, reject) => {
+  body.addEventListener('contextmenu', (e) => {
+    e.preventDefault();
+
+    resolve(
+      {
+        classMessage: 'success',
+        textMessage: 'First promise was resolved',
+      }
+    );
```
VitaliyBodnar, line 23: this objects it duplicated, so it's better to create separate function with instance of this objects, and set messages as params, for instance:
```
resolve(resultText('success', 'First promise was resolved'));
    
function getResultText(classMessage, textMessage) {
  return { classMessage, textMessage };
}
```

VitaliyBodnar (changes requested):

nazarmatsevych (approved):
🔥

PR #580, opened 2023-11-03
Solution

andrii-rykal:
[DEMO LINK](https://andrii-rykal.github.io/js_promises_practice_DOM/)


src/scripts/main.js
```diff
@@ -1 +1,63 @@
...
+
+  document.addEventListener('contextmenu', () => {
+    resolve('Second promise was resolved');
+  });
+});
+
+const thirdPromise = new Promise(resolve => {
+  document.addEventListener('click', () => {
+    document.addEventListener('contextmenu', (e) => {
+      e.preventDefault();
```
loralevitska, line 44: Don't use nesting listeners.  Create two boolean variables (falsy) instead of this and then resolve if both variables changed value (truthy)

loralevitska (changes requested):
Almost done) Let's improve

VitaliyBodnar (approved):
Good job!

PR #587, opened 2023-11-21
add task solution

kate-prakofyeva:
tests don't work
npm start don't work

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,49 @@
 'use strict';
+
+const body = document.querySelector('body');
+
+function handleStatus(message, className) {
+  body.insertAdjacentHTML(
+    'afterend',
+    `<div data-qa="notification" class=${className}>${message}</div>`
+  );
+}
+
+const firstPromise = new Promise((resolve, reject) => {
+  document.addEventListener('mouseup', (e) => {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1 +1,49 @@
...
+});
+
+const thirdPromise = new Promise((resolve, reject) => {
+  let left;
+
+  document.addEventListener('mouseup', (e) => {
+    if (e.button === 0) {
+      left = true;
+    } else if (e.button === 2 && left) {
+      resolve('Third promise was resolved');
```
Barosz30, line 46: Not sure about this. Like this, it only resolves on right click, doesnt it?
You could check if both left and right are true, and then resolve.
Just a thought.

kate-prakofyeva:
I changed the second promise because I read the tests and understood, that every promises should work only once in this task

Barosz30 (approved):
Very good code overall, I can see that we're dealing with expert here :P

src/scripts/main.js
kate-prakofyeva: you're right, it worked in sequence left-right, now it doesn't matter what was pressed first

JoyCoffeeAddict (approved):
Good job!

PR #589, opened 2023-11-22
Solution

zaura333:
[DEMO LINK](https://zaura333.github.io/js_promises_practice_DOM/)

Tests don't work

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,55 @@
 'use strict';
+
+const firstPromise = new Promise((resolve, reject) => {
+  document.addEventListener('click', () => {
+    resolve(`First promise was resolved`);
+  });
+  setTimeout(() => reject(new Error(`First promise was rejected`)), 3000);
+});
+const secondPromise = new Promise((resolve) => {
+  document.addEventListener('click', () => {
+    resolve(`Second promise was resolved`);
+  });
+
+  document.addEventListener('contextmenu', () => {
+    resolve(`Second promise was resolved`);
+  });
+});
+const thirdPromise = new Promise((resolve) => {
+  let clicksCount = 0;
+
+  document.addEventListener('click', handleThirdEvents);
+
+  document.addEventListener('contextmenu', handleThirdEvents);
+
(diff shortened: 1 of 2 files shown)


zaura333:
Third promise should work now, commit's from yesterday, because I fixed it then and didn't push it for some reason

src/scripts/main.js
```diff
@@ -1 +1,55 @@
...
+  document.addEventListener('contextmenu', () => {
+    resolve(`Second promise was resolved`);
+  });
+});
+const thirdPromise = new Promise((resolve) => {
+  let clicksCount = 0;
+
+  document.addEventListener('click', handleThirdEvents);
+
+  document.addEventListener('contextmenu', handleThirdEvents);
```
Patrykmclaren14: When utilizing the context menu in Event, be cautious, as it might be a trap. If you press Shift + F10 in the browser task, you'll notice that the operation is executed. Therefore, the intention is to either right-click or left-click.

You can use: 
`
addEventListener('mouseup', (e) => {
  if (e.button === 0 || e.button === 2)
`

src/scripts/main.js
```diff
@@ -1 +1,55 @@
...
+  setTimeout(() => reject(new Error(`First promise was rejected`)), 3000);
+});
+const secondPromise = new Promise((resolve) => {
+  document.addEventListener('click', () => {
+    resolve(`Second promise was resolved`);
+  });
+
+  document.addEventListener('contextmenu', () => {
+    resolve(`Second promise was resolved`);
+  });
```
Patrykmclaren14, line 16: When utilizing the context menu in Event, be cautious, as it might be a trap. If you press Shift + F10 in the browser task, you'll notice that the operation is executed. Therefore, the intention is to either right-click or left-click.

You can use:
```
addEventListener('mouseup', (e) => {
  if (e.button === 0 || e.button === 2)
```

Patrykmclaren14 (changes requested):
Everything looks just fine. you only need to change the operation of the context menu. 🙂

mbruhler (approved):
Approved

PR #590, opened 2023-11-22
Solution

mosowiecka:
[DEMO LINK](https://<mosowiecka>.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,72 @@
 'use strict';
+
+// write your code here
+const firstPromise = new Promise((resolve, reject) => {
+  document.addEventListener('click', (ev) => {
+    if (ev.button === 0) {
+      resolve('First promise was resolved');
+    }
+  });
+
+  setTimeout(() => {
+    reject(new Error('First promise was rejected'));
+  }, 3000);
+});
+
+const secondPromise = new Promise((resolve, reject) => {
+  document.addEventListener('click', (ev) => {
+    if (ev.button === 0 || ev.button === 2) {
+      resolve('Second promise was resolved');
+    }
+  });
+});
+
+let leftCnt = 0;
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1 +1,72 @@
...
+  document.addEventListener('click', (ev) => {
+    // ev.preventDefault();
+
+    if (ev.button === 0) {
+      leftCnt++;
+    }
+
+    if (ev.button === 2) {
+      rightCnt++;
+    }
```
zaura333, line 36: From what I checked, something doesn't really work with the right click here, same with the second promise. This could be perhaps dealt with by replacing 'click' with 'mouseup' or 'mousedown'

src/scripts/main.js
```diff
@@ -1 +1,72 @@
...
+      resolve('First promise was resolved');
+    }
+  });
+
+  setTimeout(() => {
+    reject(new Error('First promise was rejected'));
+  }, 3000);
+});
+
+const secondPromise = new Promise((resolve, reject) => {
```
zaura333, line 16: 'reject' isn't used here, so it can be deleted (same in line 26)

zaura333 (approved):
I can't open your demo, so I hope the suggestion about right click isn't useless, but that's the problem I spotted. Just that one thing
Overall, looks good!

mvjl000 (changes requested):
Your solution is not working as described in the readme.

On left click, both first and second promise gets resolved at once. Third promise is not getting resolved at all. Try to fix that

mpiskadlo (changes requested):
Have you pushed your recent changes? It still behaves the same way

mvjl000 (approved):

mpiskadlo (approved):

PR #600, opened 2023-12-06
solution

donizer:
[DEMO LINK](https://donizer.github.io/js_promises_practice_DOM/)

If i get it right, in this task some test can't be passed because second promise always resolves after first one resolves. So i skipped them.

```md
The firstPromise should be

    resolved with a message First promise was resolved on a left click in the document
    rejected with a message First promise was rejected in 3 seconds if not clicked

The secondPromise should be:

    resolved only on left or right click in the document with a message Second promise was resolved
    never rejected

```


src/scripts/main.js
```diff
@@ -1 +1,70 @@
...
+  .then((message) => {
+    onSuccess(message);
+  })
+  .catch((err) => {
+    onError(err);
+  });
+
+secondPromise.then((message) => {
+  onSuccess(message);
+});
```
SanyaBratashchuk, line 66:
```suggestion
  .then(onSuccess)
  .catch(onError);

secondPromise.then(onSuccess);
```

SanyaBratashchuk (approved):

PR #618, opened 2024-01-19
add task solution

pawelnowicki87:
[DEMO LINK](https://pawelnowicki87.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -68,7 +68,7 @@ document.addEventListener('DOMContentLoaded', function() {
 
     function checkBothClicks() {
       if (leftClick && rightClick) {
-        resolve('Both left and right clicks happened');
+        resolve('Third promise was resolved');
         document.removeEventListener('click', handleLeftClick);
         document.removeEventListener('contextmenu', handleRightClick);
       }
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -50,24 +50,23 @@ document.addEventListener('DOMContentLoaded', function() {
   });
 
   const thirdPromise = new Promise((resolve) => {
-    let leftClick = false;
-    let rightClick = false;
+    let leftClickElement = null;
+    let rightClickElement = null;
 
-    function handleLeftClick() {
-      leftClick = true;
+    function handleLeftClick(cli) {
+      leftClickElement = cli.target;
       checkBothClicks();
     }
 
-    function handleRightClick(events) {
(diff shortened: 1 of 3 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
This part of the task is incomplete
image
image

BogdanMaliuta (changes requested):
Hello, third promise doesnt work correctly

src/scripts/main.js
```diff
@@ -1 +1,85 @@
...
+
+    document.addEventListener('click', handleClick);
+  });
+
+  secondPromise.then((message) => {
+    updateNotification('success', message);
+  });
+
+  const thirdPromise = new Promise((resolve) => {
+    let leftClickElement = null;
```
yevhenii-pyl, line 53: You are complicating things here. 
There is basically no need for you to store the element that was clicked, you are only interested in fact that it was actually a click on the document. 

Why don't you check for something like `leftClicked ` that is firstly set to a false?

src/scripts/main.js
```diff
@@ -1 +1,85 @@
...
+    let rightClickElement = null;
+
+    function handleLeftClick(cli) {
+      leftClickElement = cli.target;
```
yevhenii-pyl, line 57: So that here you can just set it to true. Simple as that.

src/scripts/main.js
```diff
@@ -1 +1,85 @@
...
+      checkBothClicks();
+    }
+
+    function handleRightClick(cli) {
+      cli.preventDefault();
```
yevhenii-pyl, line 62: Whatever! There is no demand for you to avoid displaying contextmenu, no need to add an 'extra feature' when your basic task is not working.

src/scripts/main.js
```diff
@@ -1 +1,85 @@
...
+      rightClickElement = cli.target;
+      checkBothClicks();
+    }
+
+    function checkBothClicks() {
+      if (leftClickElement && rightClickElement
+        && leftClickElement !== rightClickElement) {
```
yevhenii-pyl, line 69: Here comes the worst part :)  What's the whole point here to check if these elements are not equal?

yevhenii-pyl (changes requested):
It has to be refactored, just follow my comments :)

yevhenii-pyl (approved):
Now we are talking. Works as a clock 🥇

PR #635, opened 2024-02-22
Solution

przwojwwp:


BogdanMaliuta (changes requested):
Your code does not work.Second promise must work on left or right click not only righ

PR #662, opened 2024-05-09
solutionV1

kgGugala:
- [DEMO LINK](https://kgGugala.github.io/js_promises_practice_DOM/)
    
    In my opinion works fine and I dont know why some test are falling

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,58 @@
+/* eslint-disable prefer-promise-reject-errors */
 'use strict';
+
+function createMessage(message, className) {
+  document.body.insertAdjacentHTML(
+    'afterbegin',
+    `
+    <div data-qa="notification" class="${className}">
+    ${message}
+    </div>
+    `,
+  );
+}
+
+const firstPromise = new Promise((resolve, reject) => {
+  document.addEventListener('click', () => {
+    resolve();
+  });
+
+  setTimeout(() => {
+    reject();
+  }, 3000);
+});
+
+const secondPromise = new Promise((resolve) => {
+  document.addEventListener('click', () => {
+    resolve();
+  });
+});
+
+const thirdPromise = new Promise((resolve) => {
+  let leftCounter = 0;
+  let rightCounter = 0;
+
+  document.addEventListener('click', () => {
+    leftCounter++;
+
+    if (leftCounter > 0 && rightCounter > 0) {
+      resolve();
+    }
+  });
+
+  document.addEventListener('contextmenu', () => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1 +1,58 @@
...
+    rightCounter++;
+
+    if (leftCounter > 0 && rightCounter > 0) {
+      resolve();
+    }
+  });
+});
+
+firstPromise.then(() => createMessage('First promise was resolved', 'success'));
+firstPromise.catch(() => createMessage('First promise was rejected', 'error'));
```
yevhenii-pyl, line 53: Let's chain those methods!

```
firstPromise
    .then
    .catch

```
Isn't chaining is the feature of promises? :)

src/scripts/main.js
```diff
@@ -1 +1,58 @@
...
+
+secondPromise.then(() =>
+  // eslint-disable-next-line prettier/prettier
```
yevhenii-pyl, line 56: Let's try NOT to disable linter rules. Isn't there a better way?

src/scripts/main.js
```diff
@@ -1 +1,58 @@
...
+
+    if (leftCounter > 0 && rightCounter > 0) {
+      resolve();
+    }
+  });
+
+  document.addEventListener('contextmenu', () => {
+    rightCounter++;
+
+    if (leftCounter > 0 && rightCounter > 0) {
```
yevhenii-pyl, line 46: Let's try to shorten this check a bit. :)

yevhenii-pyl (changes requested):
Good job! Works as it should to, but let's try to adjust a few minor things.

src/scripts/main.js
yevhenii-pyl: Had to try it locally, really, for some reason linter has a strange behaviour here. It works as it should to, only if we add .catch chaining, but it's not what we want here, so let's leave it as it is.

src/scripts/main.js
```diff
@@ -1 +1,60 @@
...
+  );
+}
+
+const firstPromise = new Promise((resolve, reject) => {
+  document.addEventListener('click', () => {
+    resolve();
+  });
+
+  setTimeout(() => {
+    reject();
```
yevhenii-pyl, line 21: reject(new Error(''));

/* eslint-disable prefer-promise-reject-errors */ => this rule should not be disabled

yevhenii-pyl (approved):
LGTM 
Test errors are probably due to `insertAdjacentHTML` method usage, DEMO works as expected.

PR #718, opened 2024-07-04
Solution

AnnBusya:
[DEMO LINK](https://AnnBusya.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,82 @@
 'use strict';
+'use strict';
+
+const handleSuccess = (message) => {
+  const div = document.createElement('div');
+
+  div.setAttribute('data-qa', 'notification');
+  div.classList.add('success');
+  div.textContent = `${message} promise was resolved`;
+  document.body.appendChild(div);
+};
+
+const handleError = (message) => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1 +1,82 @@
 'use strict';
+'use strict';
```
volodymyr-soltys97, line 2: Don't need to duplicate 'use strict'
```suggestion
'use strict';
```

src/scripts/main.js
```diff
@@ -1 +1,82 @@
...
+
+const handleSuccess = (message) => {
+  const div = document.createElement('div');
+
+  div.setAttribute('data-qa', 'notification');
+  div.classList.add('success');
+  div.textContent = `${message} promise was resolved`;
+  document.body.appendChild(div);
```
volodymyr-soltys97, line 10: Move this logic to the helper function and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

anastasiiavorobiova (approved):
Great work!

PR #736, opened 2024-07-18
Solution

GeorgeSavostikov:
[DEMO LINK](https://georgesavostikov.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,74 @@
 'use strict';
+
+const handleSuccess = (message) => {
+  const divElement = document.createElement('div');
+
+  divElement.dataset.qa = 'notification';
+  divElement.classList.add('success');
+  divElement.textContent = message;
+
+  document.body.append(divElement);
+};
+
+const handleError = (message) => {
+  const divElement = document.createElement('div');
+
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1 +1,74 @@
 'use strict';
+
+const handleSuccess = (message) => {
+  const divElement = document.createElement('div');
+
+  divElement.dataset.qa = 'notification';
+  divElement.classList.add('success');
+  divElement.textContent = message;
+
+  document.body.append(divElement);
```
volodymyr-soltys97, line 10: Create a helper function for this logic and use it in two places

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/scripts/main.js
```diff
@@ -1 +1,67 @@
 'use strict';
+
+const handleErrorOrSuccess = (message, type) => {
```
volodymyr-soltys97, line 3:
```suggestion
const handleMessage = (message, type) => {
```

volodymyr-soltys97 (approved):
Good job 👍

PR #741, opened 2024-07-20
add solution

smachylooo:
- [DEMO LINK](https://smachylooo.github.io/js_promises_practice_DOM/)


etojeDenys (changes requested):
demo link doesn't work, try to fix that

SanyaBratashchuk (changes requested):
Promises are not resolving on right click.

PR #776, opened 2024-08-12
solution

RomanPerepelkin:
https://RomanPerepelkin.github.io/js_promises_practice_DOM

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,97 @@
 'use strict';
+
+const body = document.querySelector('body');
+
+const promiseOne = new Promise((resolve, reject) => {
+  let click = false;
+
+  const clickHandler = (e) => {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1 +1,97 @@
...
+});
+
+promiseOne
+  .then(() => {
+    const div = document.createElement('div');
+
+    div.dataset.qa = 'notification';
+    div.classList.add('success');
+    div.innerHTML = 'First promise was resolved';
+    document.body.appendChild(div);
```
volodymyr-soltys97, line 70: Create a helper function for this logic and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

lerastarynets (approved):
Keep up the great work!

PR #782, opened 2024-08-16
promises

VadymBryvka:


src/scripts/main.js
```diff
@@ -1 +1,87 @@
 'use strict';
+
+const firstSuccessMessage = document.createElement('div');
+
+firstSuccessMessage.setAttribute('class', 'success');
+firstSuccessMessage.setAttribute('data-qa', 'notification');
+
+firstSuccessMessage.textContent = 'First promise was resolved';
```
volodymyr-soltys97, line 8: Create a helper function for this logic and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/scripts/main.js
```diff
@@ -1 +1,71 @@
 'use strict';
+
+function message(count, decide = 'resolved') {
+  const SuccessMessage = document.createElement('div');
```
volodymyr-soltys97, line 4:
```suggestion
  const successMessage = document.createElement('div');
```

volodymyr-soltys97 (approved):
Good job 👍

PR #822, opened 2024-09-11
solved

Artem-Brui:
- [DEMO LINK](https://Artem-Brui.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,60 @@
 'use strict';
+
+function buildNotification(className, text) {
+  const newDiv = document.createElement('div');
+
+  newDiv.setAttribute('data-qa', 'notification');
+  newDiv.className = className;
+  newDiv.textContent = text;
+  document.body.appendChild(newDiv);
+}
+
+const firstPromise = () => {
+  return new Promise((resolve, reject) => {
+    const callback = () => resolve('First promise was resolved');
+
+    document.addEventListener('click', callback);
+
+    setTimeout(() => reject(new Error('First promise was rejected')), 3000);
+  });
+};
+
+firstPromise()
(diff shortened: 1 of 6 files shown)


src/scripts/main.js
```diff
@@ -1 +1,60 @@
...
+
+secondPromise().then((mes) => buildNotification('success', mes));
+
+const thirdPromise = () => {
+  return new Promise((resolve, reject) => {
+    let isLeftClickWas = 0;
+    let isRightClickWas = 0;
+
+    const canResolve = () => {
+      if (isLeftClickWas > 0 && isRightClickWas > 0) {
```
VitaliyBondarenko1982:
```suggestion
      if (isLeftClick && isRightClick) {
```
it can be boolean variables. suggest better naming for it.

src/scripts/main.js
```diff
@@ -1 +1,60 @@
...
+    document.addEventListener('contextmenu', callback);
+  });
+};
+
+secondPromise().then((mes) => buildNotification('success', mes));
+
+const thirdPromise = () => {
+  return new Promise((resolve, reject) => {
+    let isLeftClickWas = 0;
+    let isRightClickWas = 0;
```
VitaliyBondarenko1982:
```suggestion
    let isLeftClick = false;
    let isRightClick = false;
```

and fix in place where it these variables changing

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest some improvements.

src/scripts/main.js
Artem-Brui: thanks, solved :)

volodymyr-soltys97 (approved):
Good job!
You need to create a new branch for your solution in the future 
image

PR #826, opened 2024-09-14
add solution

lerika7:
- [DEMO LINK](https://lerika7.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,63 @@
 'use strict';
+
+function showNotification(isResolved, promiseNumber) {
+  const message = document.createElement('div');
+
+  message.setAttribute('data-qa', 'notification');
+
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1 +1,63 @@
...
+    message.classList.add('error');
+    message.textContent = `${promiseNumber} promise was rejected`;
+  }
+
+  document.body.append(message);
+}
+
+const leftClick = new Promise((resolve) => {
+  document.addEventListener('click', () => {
+    resolve('Third promise was resolved');
```
TarasHoliuk, line 22: FYI: you don't use this value, so you could do just `resolve()` - without arguments

TarasHoliuk (approved):
Well done! Approved ✔️

PR #828, opened 2024-09-15
solution

pasha1932:
- [DEMO LINK](https://pasha1932.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -43,24 +43,23 @@ const thirdPromis = new Promise((resolve) => {
   });
 });
 
-const successHandler = (dataFromResolve) => {
+function appendMessage(data, hasError) {
   const div = document.createElement('div');
 
   div.setAttribute('data-qa', 'notification');
-  div.classList.add('success');
-  div.innerHTML = dataFromResolve;
+
+  if (!hasError) {
+    div.classList.add('success');
+  } else {
+    div.classList.add('error');
+  }
+
+  div.innerHTML = data;
   document.body.append(div);
-};
+}
 
-const errorHandler = (dataFromReject) => {
-  const div = document.createElement('div');
-
-  div.setAttribute('data-qa', 'notification');
-  div.classList.add('error');
-  div.innerHTML = dataFromReject;
-  document.body.append(div);
-};
-
-firstPromise.then(successHandler).catch(errorHandler);
-secondPromis.then(successHandler);
-thirdPromis.then(successHandler);
+firstPromise
+  .then((message) => appendMessage(message))
+  .catch((error) => appendMessage(error));
+secondPromis.then((message) => appendMessage(message));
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1 +1,66 @@
...
+  });
+});
+
+const successHandler = (dataFromResolve) => {
+  const div = document.createElement('div');
+
+  div.setAttribute('data-qa', 'notification');
+  div.classList.add('success');
+  div.innerHTML = dataFromResolve;
+  document.body.append(div);
```
volodymyr-soltys97: Create a helper function for this logic and use it everywhere

volodymyr-soltys97 (changes requested):
Good job!
Let's make your code better

src/scripts/main.js
```diff
@@ -1 +1,65 @@
...
+    div.classList.add('error');
+  }
+
+  div.innerHTML = data;
+  document.body.append(div);
+}
+
+firstPromise
+  .then((message) => appendMessage(message))
+  .catch((error) => appendMessage(error));
```
TarasHoliuk: You have forgotten to pass hasError here, so the error message looks like a success message now:
image

BTW, inside of the `appendMessage` function you can check: `data instanceof Error` instead of passing the second argument

src/scripts/main.js
```diff
@@ -1 +1,65 @@
...
+  } else {
+    div.classList.add('error');
+  }
+
+  div.innerHTML = data;
+  document.body.append(div);
+}
+
+firstPromise
+  .then((message) => appendMessage(message))
```
TarasHoliuk:
```suggestion
  .then(appendMessage)
```
you can write it in shorter way - additional function wrapper is redundant if `appendMessage` properly works with this one argument

TarasHoliuk (changes requested):
2 small comments

src/scripts/main.js
```diff
@@ -1 +1,63 @@
...
+    resolve('Second promise was resolved');
+  });
+
+  document.addEventListener('contextmenu', (e) => {
+    e.preventDefault();
+    resolve('Second promise was resolved');
+  });
+});
+
+const thirdPromis = new Promise((resolve) => {
```
SanyaBratashchuk:
```suggestion
const thirdPromise = new Promise((resolve) => {
```

src/scripts/main.js
```diff
@@ -1 +1,63 @@
...
+  });
+
+  if (!clicked) {
+    setTimeout(() => {
+      reject(new Error('First promise was rejected'));
+    }, 3000);
+  }
+});
+
+const secondPromis = new Promise((resolve) => {
```
SanyaBratashchuk:
```suggestion
const secondPromise = new Promise((resolve) => {
```

SanyaBratashchuk (approved):

PR #840, opened 2024-09-22
Complete Task

RomanMartseniuk:
- [DEMO LINK](https://romanmartseniuk.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,85 @@
 'use strict';
+
+const logo = document.querySelector('.logo');
+
+const promise1 = new Promise((resolve, reject) => {
+  logo.addEventListener('click', (e) => {
+    resolve(() => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1 +1,85 @@
...
+});
+
+const successHandler = (resolve) => {
+  const message = document.createElement('div');
+
+  message.className = 'success';
+  message.textContent = resolve();
+  message.dataset.qa = 'notification';
+
+  document.body.appendChild(message);
```
volodymyr-soltys97, line 70: Сreate a helper function for this logic and use it in two places

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The third promise does not appear on the page when we click anywhere on the page except the logo
image

PR #858, opened 2024-10-08
promises_practice_DOM

Yaroslav65:
- DEMO LINK  -  https://Yaroslav65.github.io/js_promises_practice_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,82 @@
 'use strict';
+
+const doc = document.querySelector('body');
+
+let leftClick = false;
+let rightClick = false;
+let firstPromiseRejected = false;
+
+const createDiv = (classN, textC) => {
+  const div = document.createElement('div');
+  div.setAttribute('data-qa', 'notification');
+  div.className = classN;
+  div.textContent = textC;
+  doc.appendChild(div);
+};
+const firstPromise = new Promise((resolve, reject) => {
+  const rejectTimer = setTimeout(() => {
+    createDiv('error', 'First promise was rejected');
+    firstPromiseRejected = true;
+    reject();
+  }, 3000);
+
+  doc.addEventListener('click', (e) => {
+    if (e.button === 0 && !firstPromiseRejected) {
+      clearTimeout(rejectTimer);
+      resolve();
+    }
+  });
+});
+
+firstPromise
+  .then(() => {
+    if (!firstPromiseRejected) {
+      createDiv('success', 'First promise was resolved');
+    }
+  })
+  .catch(() => {});
+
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1 +1,82 @@
...
+
+secondPromise.then(() => {
+  createDiv('success', 'Second promise was resolved');
+});
+
+const thirdPromise = secondPromise.then(() => {
+  return new Promise((resolve) => {
+    const checkBothClicksAndResolve = () => {
+      if (leftClick && rightClick) {
+        createDiv('success', 'Third promise was resolved');
```
etojeDenys: create div in then or catch functions, not here

src/scripts/main.js
```diff
@@ -1 +1,82 @@
...
+const createDiv = (classN, textC) => {
+  const div = document.createElement('div');
+  div.setAttribute('data-qa', 'notification');
+  div.className = classN;
+  div.textContent = textC;
+  doc.appendChild(div);
+};
+const firstPromise = new Promise((resolve, reject) => {
+  const rejectTimer = setTimeout(() => {
+    createDiv('error', 'First promise was rejected');
```
etojeDenys: fix the same everywhere

src/scripts/main.js
```diff
@@ -1 +1,82 @@
...
+  .then(() => {
+    if (!firstPromiseRejected) {
+      createDiv('success', 'First promise was resolved');
+    }
+  })
+  .catch(() => {});
+
+const secondPromise = new Promise((resolve) => {
+  const handleClick = (e) => {
+    if (!firstPromiseRejected) {
```
etojeDenys: why do we need all this logic with firstPromiseRejected, looks redundant for me

src/scripts/main.js
```diff
@@ -1 +1,82 @@
...
+    e.preventDefault();
+    handleClick(e);
+  });
+});
+
+secondPromise.then(() => {
+  createDiv('success', 'Second promise was resolved');
+});
+
+const thirdPromise = secondPromise.then(() => {
```
etojeDenys: it should be new promise instead of secondPromise.then

etojeDenys (changes requested):
feel free to ask for some help in this task in the chat

src/scripts/main.js
Yaroslav65: I added this logic so that when pressing "for example" -  the right mouse button, if the left mouse button had already been pressed, the message from the second promise would not be displayed again and this message would not be overlapped by the third message from the last promise

volodymyr-soltys97 (approved):
Great work! 🔥

PR #893, opened 2024-11-05
Task solution

Dominik-Dudek-code:
- [DEMO LINK](https://Dominik-Dudek-code.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,60 @@
 'use strict';
+
+const firstPromise = new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


Dominik-Dudek-code:
The tests only accept the contextmenu option, in the case of auxclick the tests are not passed, even though this solution seems more correct, an example is that on my keyboard I have a button that opens the contextmenu, which counts the promise call, even if it was not called with the mouse, and in the case of auxclick, only the right button works, no exceptions!

danon321 (approved):
Najs, fajny kod

PR #1060, opened 2025-03-08
add solution

SaXoNG:
[DEMO LINK](https://SaXoNG.github.io/js_promises_practice_DOM/)


src/scripts/main.js
```diff
@@ -1 +1,72 @@
 'use strict';
+
+function messageContent(type, text) {
```
volodymyr-soltys97, line 3:
```suggestion
const messageContent = (type, text) => {
```

volodymyr-soltys97 (approved):
Good job! 👍

PR #1078, opened 2025-03-22
add task solution

hetskoweb:
[DEMO LINK](https://hetskoweb.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -27,19 +27,17 @@ const secondPromise = new Promise((resolve, reject) => {
   });
 });
 const thirdPromise = new Promise((resolve, reject) => {
-  let firstButtonPressed = false;
+  let leftPressed = false;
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1 +1,73 @@
...
+  });
+
+  setTimeout(() => {
+    if (!isResolved) {
+      reject(new Error('First promise was rejected'));
+    }
+  }, 3000);
+});
+const secondPromise = new Promise((resolve, reject) => {
+  document.addEventListener('mousedown', (e) => {
```
volodymyr-soltys97, line 23: Don't need to use mousedown event, need to use only click and contextmenu events

volodymyr-soltys97 (changes requested):
Good job 👍 
Try to fix the failed tests, if you need some help feel free in the fe_chat

PR #1091, opened 2025-04-03
solution

xxxKoRoBoKxxx:
[DEMO LINK](https://xxxkorobokxxx.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -35,7 +35,7 @@ function showNotification(messageText, isError = false) {
   const notification = document.createElement('div');
 
   notification.setAttribute('data-qa', 'notification');
-  notification.className = isError ? ' error' : 'success';
+  notification.className = isError ? 'error' : 'success';
   notification.innerText = messageText;
   document.querySelector('body').append(notification);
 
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1 +1,69 @@
...
+  document.addEventListener('contextmenu', (event) => {
+    event.preventDefault();
+
+    resolve('Second promise was resolved');
+  });
+});
+
+const thirdPromise = new Promise((resolve) => {
+  document.addEventListener('mousedown', (event) => {
+    if (event.buttons === 3) {
```
etojeDenys, line 28: image
better to fix that

src/scripts/main.js
```diff
@@ -1 +1,69 @@
...
+
+  setTimeout(() => {
+    notification.style.display = 'none';
+  }, 3000);
+}
+
+firstPromise
+  .then((message) => {
+    showNotification(message);
+  })
```
etojeDenys, line 50:
```suggestion
  .then(showNotification)
```

etojeDenys (changes requested):

src/scripts/main.js
xxxKoRoBoKxxx: i dont understand why i must separate it into 2 eventlisteners, when event.buttons return 3 only when both buttons are clicked. His fix sounds irrationally

PR #1121, opened 2025-05-06
add solution

George-Buzhenitsa:
https://George-Buzhenitsa.github.io/js_promises_practice_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -5,25 +5,31 @@ const body = document.body;
 
 function firstPromise() {
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1 +1,86 @@
...
+  });
+}
+
+firstPromise()
+  .then((value) => {
+    const div = document.createElement('div');
+
+    div.setAttribute('data-qa', 'notification');
+    div.textContent = value;
+    body.append(div);
```
etojeDenys, line 62: it's better to create a helper function and move this logic there

etojeDenys (approved):

PR #1194, opened 2025-07-28
Solution

Adamur97:


DmytroOleksiiuk (changes requested):
Please submit the file with your solution, I and AI don't see any of them submitted. 
Now you sent only the 'technical files'.

EwaKobiela (changes requested):
please remove unnecesary files (eg. "tatus" file)

PR #1350, opened 2026-02-16
added solution

MykolaFatkullin:


MykolaFatkullin:
@mateacademy-ai-mentor If I add string instead of Error, stay lint will throw error

PR #1391, opened 2026-04-14
fix line endings

Adimate123:


tbsikora (changes requested):
Rejected as per Dmytro's request

